PR #1, opened 2017-01-11
One Piece, WWHD+missing filter

Darkemaste:
Added One Piece and WWHD graphic packs, updated existing WWHD packs with missing filter.


RHY3756547:
You can apply multiple rule sets at once. That is, you can have one folder for reduced bloom, and activate the re-scaling ones separately. Would be much better than having a ton of different packs.

Darkemaste:
I see, thanks.

slashiee:
okay, a few things: 
- formatting in one piece seems fucked, there's spaces in between commands
- i'm not gonna accept the no bloom versions of wind waker, i don't really want nintendo to be on my ass just because i'm including one of their shaders
- you need to use the proper game name for one piece for the readme
other than that, it seems good

PR #2, opened 2017-01-11
[MK8] Add "/meta/"less id, fullres alpha, MKTV, fix race end graphic

RHY3756547:
4K alpha (i'm using a 1080p monitor):
image

4K in MKTV:
image

Excludes the format used by the backgrounds used at the end of VS Mode/Grand Prix, so they don't appear black.

Also ends up removing the gamepad rescale. I think that should be in a separate pack, as in most cases users won't be looking at the gamepad in this game, and rendering it at a higher res would just burn GPU cycles.

### DIFF
--- a/MK8_1080p/rules.txt
+++ b/MK8_1080p/rules.txt
@@ -1,16 +1,43 @@
 [Definition]
-titleIds = 000500001010ec00,0005000e1010ed00,0005000e1010eb00
-name = "Mario Kart 8 - 1080p"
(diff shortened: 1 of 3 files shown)


slashiee:
Thinking about it, removing gamepad scaling would be for the better. I'll make a separate commit to this repo that gets rid of the gamepad scaling.
Also, I would probably show examples of issues that are fixed, if possible.

PR #3, opened 2017-01-12
[Graphics Packs] Add graphic packs for various titles.

Darkemaste:
4K screenshots for a few included titles (I'm using a 1440p monitor)
opuwr1
ttt2
stz1
ftsp
ssprr

Note: Star Fox Zero uses custom resolutions for both the TV and Gamepad, I've already fixed the TV resolution but not the gamepad resolution. (Both TV resolutions are required!) 
Also possibly requires a filter/rule for objects in the loading screen and  pause screen when using higher resolutions.


slashiee:
Do you think you can combine all the commits into one single commit? There's way too many here. Other than that, it seems okay to me.

slashiee:
Nevermind, I figured out how to squash them.

PR #8, opened 2017-01-17
MK8 2P & 4P MP Res

Darkemaste:
Mario Kart 8 split-screen magic revealed-

4 Player Multiplayer at native resolution
mk84p-native

4 Player Multiplayer at half-width 1440p (2K)
mk84p-2k

4 Player Multiplayer at half-width 2160p (4K)
mk84p-4k

### DIFF
--- a/Quality/MK8_1080p/rules.txt
+++ b/Quality/MK8_1080p/rules.txt
@@ -18,6 +18,22 @@ formatsExcluded = 0x41A # exclude obvious textures
 overwriteWidth = 960
 overwriteHeight = 540
 
+# Multiplayer-Mode Resolution (2p)
+[TextureRedefine]
+width = 640
+height = 720
+formatsExcluded = 0x41A # exclude obvious textures
+overwriteWidth = 960
+overwriteHeight = 1080
+
+# Multiplayer-Mode Resolution (4p)
+[TextureRedefine]
+width = 640
+height = 360
+formatsExcluded = 0x41A # exclude obvious textures
+overwriteWidth = 960
+overwriteHeight = 1080
+
 # MKTV screen
 [TextureRedefine]
 width = 960
--- a/Quality/MK8_1080p/rules.txt
+++ b/Quality/MK8_1080p/rules.txt
@@ -26,14 +26,6 @@ formatsExcluded = 0x41A # exclude obvious textures
 overwriteWidth = 960
 overwriteHeight = 1080
 
-# Multiplayer-Mode Resolution (4p)
-[TextureRedefine]
-width = 640
-height = 360
-formatsExcluded = 0x41A # exclude obvious textures
-overwriteWidth = 960
-overwriteHeight = 1080
-
 # MKTV screen
 [TextureRedefine]
 width = 960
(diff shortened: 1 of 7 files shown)


Quality/MK8_1080p/rules.txt
```diff
@@ -18,6 +18,22 @@ formatsExcluded = 0x41A # exclude obvious textures
...
+overwriteWidth = 960
+overwriteHeight = 1080
+
+# Multiplayer-Mode Resolution (4p)
+[TextureRedefine]
+width = 640
+height = 360
+formatsExcluded = 0x41A # exclude obvious textures
+overwriteWidth = 960
+overwriteHeight = 1080
```
RHY3756547: Shouldn't this be lower than the width? Should be half.

ALSO: this conflicts with the half alpha res rule. Here are two possibilities

- Assume half-alpha rule will deal with 4p render target, but at great performance cost (at 4k, you'd be pushing 4x 4k targets)
- Make half res target rescale to half of the target resolution. Will make single player use half res alpha like the original game, but at least it won't be half of 720p.
- Different graphics packs for 4p multiplayer and single player.

Have these sizes been verified with a graphics debugger? I thought they might actually be uneven due to the black bars.

Darkemaste: I didn't notice black bars or anything of the sort after extensively testing 1440p both 4K 2P and 4P multiplayer tbh, though I probably should of included screenshots as well. (will add a few soon) 
As far as half-width and normal height for 4P multiplayer it seems thats what Nintendo was already doing for 720p in perfstudio so it only seemed logical to do the same half width-normal res for other target resolutions. 
I didn't notice the conflict specifically for 4P MP and half-res alpha though, I'll look into possible solutions.

theboy181:
Will these work with the updated mk8 v4.1? Gfxpacks don't appear to be working. I think nintendo changed the internal resolution or cemu is broken.

Quality/MK8_1080p/rules.txt
RHY3756547: Your current screenshots are supersampling vertical due to the incorrect height setting I pointed out. You can see it on the 1440p screenshots.

I'm not sure how overwriting rules work in different files, it might help if exzap added some kind of order number to assign to graphics packs so that certain ones could overwrite others on demand. (eg. a multiplayer texture rule would overwrite full-res alpha one in the normal texture pack when enabled)

RHY3756547:
For graphics packs to work your game ID must be in the title ids section. If your dump is missing the meta folder, it will look something like:

ffffffff85887bc1

You can see this number in the log file cemu generates, if you have logging enabled. Add that to your graphics pack.

Quality/MK8_1080p/rules.txt
Darkemaste: Alright I see what happened with the "conflict", I wasn't using the current half-res alpha rule in my own Cemu graphic pack folder when I was trying to fix MK8 4 player multiplayer resolution. Apparently my "4P multiplayer" rule isn't needed as the existing half-res alpha rule also fixes the same issue. 
Tbh I wanted to keep the half-width normal height format(?) used for 2 player splitscreen/multiplayer in 4 player splitscreen/multiplayer but I'll just make a higher res graphic pack instead and avoid any potential half-res issues all together.

slashiee:
Everything seems good here. Merged.

PR #21, opened 2017-01-22
XCX reduced bloom

getdls:
Reduced bloom for XCX (same approach as zelda)
Only verified on Nvidia

### DIFF
--- a/Enhancements/XCX_reducedBloom/2a706eb8eef36208_00000000000003ca_ps_source.txt
+++ b/Enhancements/XCX_reducedBloom/2a706eb8eef36208_00000000000003ca_ps_source.txt
@@ -1,0 +1,85 @@
+#version 400
+#extension GL_ARB_texture_gather : enable
+
+const float bloomFactor = 0.7; // bloom strength
+
+// shader 2a706eb8eef36208
+uniform ivec4 uf_remappedPS[1];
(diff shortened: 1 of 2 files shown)


slashiee:
No shader mods allowed. I don't want my ass fucked by Nintendo.

slashiee:
I could possibly make a No Bloom out of this, though.

```
// Tex0 addr 0xf470a000 res 1280x720x1 dim 1 tm: 4 format 0816 compSel: 0 1 2 5 mipView: 0x0 (num 0x1) sliceView: 0x0 (num 0x1) Sampler0 ClampX/Y/Z: 2 2 2 border: 0
// Tex1 addr 0x25e75000 res 16x16x16 dim 2 tm: 7 format 001a compSel: 0 1 2 3 mipView: 0x0 (num 0x1) sliceView: 0x0 (num 0x10) Sampler1 ClampX/Y/Z: 2 2 2 border: 0
```

getdls:
Would be it anything but a on/off solution? There's also reflection/"spectacular lights" 
Either way won't the whole vs/ps copyright problem go away with diff patching? I thought that was being looked into. 

```
#[TextureRedefine]# Probe beacon glow. secondary fill glow general, doesn't scale well		
#width =	80	#
#height =	46	#
#formatsExcluded = 0x816  #fix probe glow		
#overwriteWidth =	240	#
#overwriteHeight =	135	#

```
```
#[TextureRedefine]# secondary glow general (quarter res dof wider), doesn't scale well		
#width =	160	#
#height =	90	#
#formatsExcluded = 0x816  #fix probe glow over sharpen		
#overwriteWidth =	480	#
#overwriteHeight =	270	#
```

slashiee:
Hmm, I might allow shader mods again. I feel like nothing significant might really happen.

getdls:
You could do a shader mod fork. If there's a shakedown notice, it's less likely cemu_graphic_packs gets locked.

PR #26, opened 2017-02-03
Added Super Mario Maker and change color splash name

Gabryxx7:
Added Super Mario Maker 1080p, 1440p, 2160p and Gamepad as well. There
are some issues though. I also renamed color splash to Paper Mario Color
Splash and added tv and gamepad comments.


slashiee:
Super Mario Maker was removed due to the fact that NSMBU mode wasn't working. It will be added back once we can figure out how to fix this issue. The name change for Color Splash is pretty unnecessary as well. I will fix the issue with the gamepad names though.

PR #44, opened 2017-04-11
Breath of the Wild HiResShadows Fix

Dat1AZNBanana:
fix from info from the megathread on reddit


slashiee:
I don't see any changes.

PR #48, opened 2017-04-30
[Multiple] Get rid of 0x001 FMV Exclude

slashiee:
[Cemuhook](http://sshnuke.net/cemuhook/) uses -16 width now for MP4 videos, plus it appears that it's 1280 check was removed from it as well. Haven't tested much games yet, will need someone to test games that I don't have on me

Games that don't require 0x001:
Pikmin 3
Pokken Tournament
Bayonetta 2

Games that do require 0x001:
Mario & Sonic at Sochi 2014
Paper Mario Color Splash

Games that don't break because of 0x001:
Sonic Boom: Rise of Lyric
Star Fox Zero
Star Fox Guard
Tekken Tag Tournament 2
Bayonetta

Games not tested:
Scribblenauts Unlimited
Scribblenauts Unmasked

### DIFF
--- a/Enthusiast/Bayonetta2_2880p/rules.txt
+++ b/Enthusiast/Bayonetta2_2880p/rules.txt
@@ -5,7 +5,6 @@ name = "Bayonetta 1 + 2 - 2880p (5K)"
 [TextureRedefine] # tv
 width = 1280
 height = 720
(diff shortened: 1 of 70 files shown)


slashiee:
actually eh.. we don't really need this

Issue #49, opened 2017-05-02
Folder Structure


Sureiya:
I know the verbosely named folders are nice to have, but I feel like it would be way more useful to have single folder `graphicPacks`. 

Instead the graphics pack category could be in the actual pack folder name, in the graphicPack name field, or categorized in a table in the docs.

It would be a lot simpler and easier to update if we could just drag the whole folder in, or for those of us comfortable with git, just do a git pull.

If no one agrees, I'll work on a PR build script or something to move them.

slashiee:
This could probably be addressed, with wiki pages telling what packs are part of what. I don't know how @Darkemaste or @getdls think about this, though.

slashiee:
I just created the Wiki part, but I don't know about the other's opinions yet. https://github.com/slashiee/cemu_graphic_packs/wiki

slashiee:
I've created a PR that would suit your needs at #50, but I'm still waiting on the others before doing anything else related to this.

getdls:
I think it's an improvement, and it doesn't require any more work from our side. That and future texture packs can probably be a flat "main folder" structure.

Enhancements,  I'd say the majority want better shadows that so keeping those at x2  and sticking them in the same folder is a reasonable compromise. It's less work to un-chek the box than to find it on git and copy .

Workarounds, just give them a really descriptive name and stick them  in the main folder, from the number of "AMD is broken questions" ppl don't know where to look for them. 

I'm still not sold on low res packs, has anyone with a slow card benched marked them against no gfx pack active at all? But I'm not that annoyed so do as you please there.  

I'd keep fun and AA removal in a separate folder, too many questions on reddit why they should be used.

Any other suggestions?

slashiee:
@getdls I've made several pack names more descriptive according to what you have said. 8a2761e9ecf786df257693cc4c69914fee1e4384

slashiee:
I've removed 720p improved packs and Fun packs altogether. I'm unsure about Performance packs though

Darkemaste:
I don't really think this is needed tbh, I think even if we did this people would then complain about the graphic pack window in Cemu being too cluttered. (Currently if you use a single category you only see 3-4 resolutions for each supported title in Cemu's graphic packs window, if you combined them all you'd then see 10-12 resolutions for each supported title instead.)

[...]

PR #50, opened 2017-05-03
Folder Structure

slashiee:
This pull request takes care of issue #49. The general idea of the issue is to put everything into a single folder for an easy drag and drop, allowing for easy updates and prevents questions that shouldn't even be there in the first place. I'm still waiting on the others for their opinion on the issue though. If this does happen to be merged, then people will be redirected to the wiki pages telling what packs are part of each 
section instead of being taken to the folders themselves.


slashiee:
Closing due to no recent responses about the issue. I may open this back up in the future.

PR #52, opened 2017-05-23
1.8.0

slashiee:
Current changes for 1.8.0, will merge once public releases

Keep in mind:
Any game that does not boot in 1.8.0 does NOT get "version = 2"


VictorHartman:
BreathOfTheWild_1080p isn't working on my pc. However 1440p, 1800p and 2160p do work.

Issue #74, opened 2017-07-31
UW graphic_pack patches don´t work right.


LoneGermanWanderer:
Cemu 1.8.2b + CemuHook 0.5.3.1 + Win10x64
i7 7700K@4.5GHz + 1060gtx6GB + 32GB sysRAM

The SuperMario3DWorld_1080p_UW patch don´t work with my copy of 
Super Mario 3D World (Cache/SaveDirNr: 7bba927a, GameID: ARDP01).

The MK8_1080pUW patch have the wrong aspect ratio in Multiplayer.
![mariokart](https://user-images.githubusercontent.com/30589348/28775279-64439d38-75f1-11e7-8ee1-a72dc94d4033.jpg)

slashiee:
Super Mario 3D World might not be the same RPX for all regions, causing it to not work for you.

Nothing can be really done about Mario Kart 8 though I believe, because I don't think there's a way to detect that it's multiplayer in patches.txt

Xalphenos:
Looking in to sm3dw right now.

LoneGermanWanderer:
Cool thx^^
i have found a temporary solution for the mario kart multiplayer.
![mariokart2](https://user-images.githubusercontent.com/30589348/28798313-67d8cdf8-7644-11e7-95f4-a66504eb3802.jpg)

Xalphenos:
For now change the modulematches line to "moduleMatches = 0xD2308838" for sm3dw.  Codes the same but the module checksum is different.  Got to test JP, and all the captain toads before adding a new patches.txt.

PR #76, opened 2017-08-05
Updating MK8 UW patches

Xalphenos:
Altered the way the aspect ratio is changed for ultrawide packs.  This method corrects the aspect ratio for two player split screen and 3d models on the multiplayer select screen.  Though the latter was close enough that most people probably didn't notice.   This addresses the other half of issue #74 even though it's already closed.

Also separated v4 and v4.1 due to one of the rodata constants being 0x10 off.  Not sure if that could have caused any problems or not.

### DIFF
--- a/Quality/MK8_1080pUW/patches.txt
+++ b/Quality/MK8_1080pUW/patches.txt
@@ -59,7 +59,27 @@ _scaleAddr = 0x00000000
 0x024642E8 = bla _scaleAspect
 
 [MK8AspectVer4]
-moduleMatches = 0x1D398493, 0xD09700CE
+moduleMatches = 0x1D398493
+0x100C359C = .float 2.370
+0x10121D30 = .float 2.370
+
+#aspect scaling
+codeCaveSize = 0x18
+
+_scaleAspect = 0x00000004
+0x00000000 = .float 1.333333
+_scaleAddr = 0x00000000
+0x00000004 = fdivs f9, f13, f12
+0x00000008 = lis r7, _scaleAddr@ha
+0x0000000C = lfs f12, _scaleAddr@l(r7)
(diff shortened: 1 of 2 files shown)


LoneGermanWanderer:
Mario Kart Version 4.1(Cache/SaveDirNr: 9708a1d3, GameID: AMKP01)
The new patches.txt woks only if i put v4 and v4.1 together.
Like this:

> [MK8AspectVer4]
> moduleMatches = 0x1D398493, **0xD09700CE**
> 0x100C359C = .float 2.370
> 0x10121D30 = .float 2.370
> 
> #aspect scaling
> codeCaveSize = 0x18
> 
> _scaleAspect = 0x00000004
> 0x00000000 = .float 1.333333
> _scaleAddr = 0x00000000
> 0x00000004 = fdivs f9, f13, f12
> 0x00000008 = lis r7, _scaleAddr@ha
> 0x0000000C = lfs f12, _scaleAddr@l(r7)
> 0x00000010 = fmuls f7, f9, f12
> 0x00000014 = blr
> 
> #replace math with branch
> 0x024AEBEC = bla _scaleAspect
>

Xalphenos:
Ok I can duplicate your issue and it's not what it seems.  Apparently [MK8AspectVer4.1] isn't a valid group name.  The better solution is to rename [MK8AspectVer4.1] to [MK8AspectVer41]  Either that or use the v33, v48, etc naming convention.  Which is probably what I'll do.

What you did works fine but it overwrites another place in rodata, 0x10121D30, with an aspect ratio.   It looks like a pointer but it may be useless garbage and not matter.   The correct offset for 4.1 is 0x10121E30.  At the last minute I decided not to chance it and separate v4 and v4.1.

PR #91, opened 2017-10-03
fix camera rune dof blur using the same method as zl focus

NAVras-Z:
btw dof turning dark is because a certain 320x180 format not scaled, which isn't fixed in this (or other) shader, as i hardly understand it...

### DIFF
--- a/Enthusiast/BreathOfTheWild_2880p/c92c1c4c0a2fb839_0000000000001e49_ps.txt
+++ b/Enthusiast/BreathOfTheWild_2880p/c92c1c4c0a2fb839_0000000000001e49_ps.txt
@@ -133,7 +133,7 @@ if( activeMaskStackC[2] == true ) {
 vec2 R9f = vec2((passParameterSem2.x + passParameterSem2.z) , (passParameterSem2.y + passParameterSem2.w)); //center point
 ivec2 ires = textureSize(textureUnitPS1,0);
 vec2 res = vec2( float(ires.x), float(ires.y) );
-int r = int(floor(2.0 / uf_fragCoordScale.y + 0.5));
+int r = int(floor(1.0 / uf_fragCoordScale.y + 0.5));
 vec3 R10f = vec3(0.0);
 float count = 0.0;
 for( int x=-r; x<=r; x++ ) {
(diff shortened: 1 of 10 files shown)


NAVras-Z:
wait... sth's broken, need to fix it first

Crementif:
Good job fixing it dude! Looks great. Also, I think that Cemu 1.10.0 might fix the blur layer resize, like it did for Mario Kart 8's blur fix. Probably unrelevant tho. 👍

NAVras-Z:
thx, i fixed the tint but not brightness, cb0 is edited in my last pull request, probably hardcoded somewhere else (or this one as when i made it pass untouched texture and i got full bright)
later i found it expects itself to be exactly q-res, the format is 320x180 0x186, same as bloom sadly.
when i upscale it:
image

slashiee:
Yeah, we just need someone to figure out how to fix the bloom

NAVras-Z:
https://cdn.discordapp.com/attachments/292733452590120961/365058448611737611/image.png
https://cdn.discordapp.com/attachments/292733452590120961/365058940582625290/image.png
bloom uses mips, and texture redefine doesn't take care of mips (so size mismatched), need to scale them all (or the one that is used in this case)

alexkiri:
I think the shader can be adjusted to always use the level0 mip, if the rest of the mips have invalid data. This will have some side effects, but it's better than nothing.

NAVras-Z:
welp my english sucks xD
i mean it's fixed by scaling mipmaps it uses.

alexkiri:
I think that mipmaps are supposed to be generated automatically when uploading the texture. At least that is how I remember it was done in OpenGL, but you had to specify that it should generate mipmaps. The fact that the rescaled texture does not have mipmaps, could be a bug in Cemu.

NAVras-Z:
it does have, just size mismatched so become "undefined data", probably another thing on devs' to-do list.
anyway, missing heatwave texture when scaling half-res to full-res shares the same root cause as bloom.
normal
wave
broken, 2nd mip have the same as the 1st(0)
heat

i made some observation, when broken the heatwave is actually working, just the main texture passed to it is black so it turns into what we saw, those weird pattern on image is the 32x32 heatwave texture itself.
also getdls mentioned it turns normal after a few seconds, i had the same when i broke dof, i guess it's actually a fail safe of the game.

NAVras-Z:
oh well i guessed wrong, thx joel for pointing out
the game wants 320x180, not 1st mipmap, i get bloom on 1440 because 160x90 is scaled to it.

slashiee:
Is this good to merge?

NAVras-Z:
i think so?
pull request itself doesn't include any bloom related thing

PR #93, opened 2017-10-06
Antialiasing restoration

alexkiri:
This restores the original Antialiasing, rather than replacing the whole shader with comletely different implementation, and also, rather than removing it completely.
This is achieved by replacing the hardcoded resolution info passed to the shader in uf_remappedPS with a size taken from the texture unit - textureSize(textureUnitPS0, 0);

Screenshots. All are in 1440p:
AABroken
aabroken

AARemoved
aaremoval

AARestored
aarestored


BreakPoints:
Why is "Contrasty" enabled by default now? That wasn't mentioned in the pull request

alexkiri:
It thought pretty much everyone wanted it. I'll disable it

BreakPoints:
I was not really a fan of combining the different graphics packs to the resolution pack from a user perceptive in the first place. It's not immediately obvious that the resolution pack changes more than just resolution and adding related fixes. Thanks for following up!

PR #106, opened 2017-10-12
Correct reflections

alexkiri:
What the title says.
reflno
reflyes

Also included some code cleanup for the AARemoval shaders, to make it easier to add them to other custom resolutions, because they are not resolution independent.


alexkiri:
Changes already committed to master

PR #107, opened 2017-10-14
remove blur shaders

NAVras-Z:
https://github.com/slashiee/cemu_graphic_packs/issues/38#issuecomment-336607945
They probably drop performance:
http://rastergrid.com/blog/2010/09/efficient-gaussian-blur-with-linear-sampling/#comment-453
Might also because more taps at larger res is bad, the largest seen in game is 7(splat)/5(botw), whereas it reaches 19/25 at 5k. And the one pass blur is even more inefficient.


alexkiri:
wait... like, all of them?

Crementif:
Probably should make them optional then, some people can push their stuff more couldn't they?

alexkiri:
A lot of work was done for implementing these. Especially finding them took considerable effort

alexkiri:
But yes, I see what you mean, the performance impact is surely a sensible issue for regular users. I agree to remove anything blur related, or, at least, make a separate BlurCorrection pack, with a side note that it might impact performance considerably. and while at it, I think you should add these shaders in the bucket as well:
f0a615c0f6a37661_0000000000000079_ps
45d85f1d25e7d0de_0000000000000079_ps
They seem copy pasted, the reflection on is for sure

alexkiri:
All in all, if the camera dof effect look ok-ish, and it doesn't impact performance, I think people can live with this:
image

Crementif:
Okay, I've talked to Kiri and the plan is that we're removing all the blur shaders for now to create a new branch with 'fixed' shaders that are less intensive (maybe something with the vertex shaders). It probably won't look as good but would remain radius and remove the grid look.

Crementif:
As of the camera, https://github.com/slashiee/cemu_graphic_packs/issues/38#issuecomment-336607945 indicates that it also is a fps hog so probably should stay removed.

PR #108, opened 2017-10-14
Add modification packs for Xenoblade Chronicles X

lasyan3:
Add 2 new packs. One that increases collectibles range, other that allow customization of drop ratio for enemies loot.

### DIFF
--- a/Modifications/XCX_CollectiblesRange/patches.txt
+++ b/Modifications/XCX_CollectiblesRange/patches.txt
@@ -1,0 +1,59 @@
+[XCX_COLLECTIBLESRANGE] 
+moduleMatches = 0xF882D5CF
+
+codeCaveSize = 0xC0
+
+_itemRangeInner = 0x00000000
+0x00000000 = lis r5, 0x4250
+0x00000004 = addi r5, r5, 0x0000
+0x00000008 = stw r5, 0x4(r23)
+0x0000000C = lfs f10, 0x4(r23)
+0x00000010 = fadd f10, f9, f10
+0x00000014 = blr
+
+_itemHeightInner = 0x00000020
+0x00000020 = lis r5, 0x4250
+0x00000024 = addi r5, r5, 0x0000
+0x00000028 = stw r5, 0x4(r23)
+0x0000002C = lfs f12, 0x4(r23)
+0x00000030 = fadd f12, f10, f12
+0x00000034 = blr
+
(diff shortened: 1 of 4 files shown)


Crementif:
Don't have the game, but maybe someone else can review it?

lasyan3:
Hi,

Just to explain a bit more what those 2 packs does (I'm planning to post the same text on cemu forum once the packs are online):
- **Collectibles range**: this pack strongly increases the distance from which you loot the collectibles. By default, you'll get them even before you're close enough to see the blue gem (distance increased by 52.0). There are 3 kind of distances in the game, each with a value of "Range" and "Height": "Inner" (which means when you are at foot), "Doll" (when you have a Skell at land) and "Flight" (which I believe is when you flight, generally in a Skell). Since I don't have a Skell yet in my current play, I would be grateful if someone could test the pack with its Skell (or even better, share with me its savegame). Right now I don't change the random selection of items (rare items are still rare).
- **Custom drop ratio**: this pack allows you to set a minimum drop ratio of **materials**. I insist on that: only materials seem concerned (if you get a gold chest, you'll still only get one or two weapon/armor, not all of them the enemy can drop). The default minimum ratio is 20%, but can easily be tweaked (if you set to always 100%, as long as the enemy drop a chest you'll get all materials from him). Second optional setting is to force the dropped chest (gold, silver or bronze). This setting was mainly for testing purpose, but some may find it interesting.

slashiee:
What version of XCX does this use?

Crementif:
After inspection, it seems that it only works without the updates. Did you use any updates?

lasyan3:
I'm using 1.0.1E (module checksum is 0xF882D5CF). I think I have the DLC but gonna check.

Crementif:
If I'm not wrong this won't work. The memory patch addresses are different for each module.

lasyan3:
I've been able to get my hand on a 1.0.2U and this works for me as long as I add the module checksum 0x30B6E091 to the list of "moduleMatches". Still wait for someone else to confirm if possible.

slashiee:
I need a save file where I can test these packs easily

lasyan3:
Here: https://drive.google.com/file/d/0B5p7e1g-NHIvNU4tNjVvZ0RXbUU/view
(taken from https://www.youtube.com/watch?v=HdFC5jDox4o)

PR #109, opened 2017-10-14
[BotW] Added fixed radius blur shaders

alexkiri:
This corrects the effects that use blur (Main menu slideshow, AO shadow, camera DOF, battle DOF, reflections, bloom)
This is done by updating the vertex shaders to calculate texture coords correctly.
This completely eliminates grid artifacts from the above effects, and has no impact on performance.
The downside is that the number of samples is constant, and in turn, the blur radius. This decreases the impact of those effects, mainly noticeable in resolutions above 1440p

Screens, both for 1440p
Current upscaled blur:
blurresuupscaled
blurupscaled
Fixed radius blur:
blurimproved
blurresuimproved


NAVras-Z:
(have to make sure every blur is upscaled by graphic pack rules)

alexkiri:
I forgot to added screenshots. I have some. Will add

NAVras-Z:
which one is which?
weird... there're stretched horizontal/vertical blur in the first ones, neither packs should produce this artifact.

alexkiri:
First 2 pics are what's currently in the main packs (with all the blur related shaders removed)
Last 2 pics are after this update

NAVras-Z:
oh ok, forgot they're removed xD

NAVras-Z:
oops still...
just guessing, I don't really understand those vertex shader... but remappedVS sometimes is involved in the calculation of gl_position, so if a texture is unscaled(mainly small ones) but remappedVS is, it could lead to some problem (but yea i haven't seen any)
zl
I tried to find an indicator of rescaling in vertex shader like the one in pixel, but whatever uf_*** is, it's both 0 (or very close to) in the blur shader...
uf_window

alexkiri:
From what I tested, these vshaders I edited have that resolution info all the time, all least in BotW 1.3.1. You normally should see 1280.0, 720.0 and some small values

NAVras-Z:
try the menu blur on zelda logo? that's where i found that, i'm on 1.3.1 as well
(yea i need to check in game

alexkiri:
Yup, I think I started with that one

alexkiri:
Wait, I think I know what you mean now. What if a vertex shader actually has correct values, and 1280, 720 are correct regardless of resolution. I am certain that the shaders I edited are for very specific functions, it's unlikely that they are reused for other functions. But there no way to tell for sure other than testing

NAVras-Z:
oh that one too xD actually i haven't thought of that
i meant a more specific one, e.g. 0b9b8f5dfa16ad58 is both used for zelda logo and q-res blur in title menu, q-res is scaled but zelda logo isn't, so when comes to the logo, remappedVS shouldn't be scaled.
but yea what happens later has too many possiblities,, only testing can tell (which surprises me that it turns out fine xD)

Crementif:
It could be that it does the upscaling by Cemu (partially) later?

NAVras-Z:
yea i guess so, if the coordinates are calculated correctly in shader it shouldn't require a uf_fragCoordScale?

Crementif:
uf_fragCoordScale is really something we shouldn't use due to the Cemu
backwards compatibility (we know this rule might not apply to some/all of
the BotW shaders) but it has different values between Cemu 1.10.0 and Cemu
1.9.1 for most games. To use this thing as a dynamic variable wouldn't be
smart.

NAVras-Z:
uf_fragCoorScale should be consistent (in theory)
image
but yea I've read your research.
For me tho, every shader that I changed got expected(?) values from it (or at least works i'm not sure) 
texturesize() returns weird values when there're mipmaps (even if there's no mip gets passed into shader at all), causing the stretched horizontal/vertical bloom.
anyway it's hacky, also history now xD passing the right value from vertex should be better 
btw @alexkiri do you consider merging this? always can update if needed

alexkiri:
Yes, I consider this should be merged. A lot of people are confused that the master pack has broken bloom effects. The people that tested this are ok with it, the only complains were about a double clouds view, that I traced to be BotW version incompatibility (version older than 1.3.0)
This solution is simple, an scalable to any custom resolutions just by changing the 2 constants on the begining of each shader

but
We must definitely bring back the old blur, the horz/vert pairs of files worked really well, and scaled liniarly for higher resolutions. The only issue was in some one-pass blurs.

NAVras-Z:
ok~
ps: some of the old blurs didn't scale linealy, I was trying to get the same look so I added more taps to them(which adapted to splatoon iirc). but only later i found it's because the downscaling and the pixel average(2x2) before it makes the blur different. so,, scaling linearly should be mathmatical correct i guess.

slashiee:
Gonna merge latest master into this branch so that it can compile appveyor with the latest changes.

NAVras-Z:
image
turns out the comment in the rastergrid belongs to the author of sweetfx himself, although I got no idea what he's talking about xD
there're 7/9/17 taps blur implemented in sweetfx's lumasharpen

NAVras-Z:
also this:
o
f
The reason why uf_fragCoordScale doesn't work when debugging, unused uniforms won't exist during runtime. So unless a shader is using it before, or adding it in nsight would not work.

alexkiri:
That discovery about uf_fragCoordScale is awesome. Makes sense to completely avoid it, then. Do you know of any other unused uniforms in the same situation?

NAVras-Z:
maybe the window** in vertex as well? it's from cemu as well, i dunno. anyway, as long as a uniform is not used in the first compiling it will be optimized out i guess.

PR #111, opened 2017-10-15
[BotW] another shader to host contrasty

NAVras-Z:
Scrolling backwards through events until I found a shader that uses main texture, seems to have sth to do with bloom but I'm not sure. Needs proper testing.


NAVras-Z:
oh yea it applies bloom in some way
org
with R1f commented
break180

alexkiri:
This is good news. We can finally keep it clean. Are you sure the shader applies changes to everything?

slashiee:
Why isn't the pack named Contrasty?

NAVras-Z:
@alexkiri I think so...?
with R0f changed to 0,0,1:
blue
@slashiee ( ͡° ͜ʖ ͡°)

NAVras-Z:
update: merge Adjustable Bloom into this Contrasty, since the original shader seems to be the final stage of bloom, also because before version2, Adjustable Bloom was part of Contrasty (in AA Removal)
original:
orgg
updated Contrasty with default value:
bloom

Crementif:
Nice, now we can have everything graphically be enabled (?) within Cemu. I'd suggest having the packs have default values (so it looks like stock BotW) to prevent people having both Clarity in their resolution pack and Contrasty enabled at the same time having to screw with the settings to change it back.

Crementif:
Got Jamie's confirmation that we're free to replace the old Contrasty with Clarity in the discord so we should probably change them all. Probably should leave this until we're sure everything is merged correctly and won't make an issue if someone would have clarity in their resolution pack.

NAVras-Z:
I have some questions adapting clarity, so it's probably better Jamie him/herself make the change.

NAVras-Z:
// no floor scale
// except saturation is used, but isn't signed a value, (0 by default?)
// fog color still black, doesn't remove anything
//defined gamma correction isn't used
//ContrastSaturationBrightness function isn't used ?looks like another crush color(tonemap)
//luminance coefficient rec.601(game) 0.2989,0.5866,0.1144 / srgb 0.2126,0.7152,0.0722
//gamma compressed before custom tone mapping
//double tone map 
//contrasty removed, only tone mapping left

NAVras-Z:
regarding bd8bb, I got the formula, but no idea why it is like that.
looks like the color is already mapped at [0,1] and then clampped at //9
start from //10 seems to be post-processing,  saturation maybe? (the reason I reverted contrasty)
//10 and later removed
nosat
//original
org

Crementif:
Sounds like you've got some really useful information here, good work.

>

NAVras-Z:
Probably not gonna get anywhere but:
RGB is texture colors, Y luminance
linear color space

sorry for the clumsy writing

NAVras-Z:
[bd8bba59e2149449_00000000000003c9_ps.txt](https://github.com/slashiee/cemu_graphic_packs/files/1406957/bd8bba59e2149449_00000000000003c9_ps.txt)
Requested by SkalFate on discord.
Keep in mind that I don't understand any of the formulas above so my assumptions are probably still wrong, or not good to modify it to begin with.

NAVras-Z:
a quick guess i thought of
guess

NAVras-Z:
eh... looks like A has sth to do with saturation (also the green tint because human eyes are more sensitive to green?)
uf_remapped.y changed from 0.18 to 1.0
green

RaverX3X:
Nice.

NAVras-Z:
oh looks like I was wrong about green tint, with it's own saturation adjust the color doesn't to green it seems
0.5:

original
orgno
could be that scene has too much green
anyway i tried to scale each color's increasement/decreasement, but ended up with more green... xD

NAVras-Z:
[bd8bba59e2149449_00000000000003c9_ps.txt](https://github.com/slashiee/cemu_graphic_packs/files/1421384/bd8bba59e2149449_00000000000003c9_ps.txt)
mess with the original saturation, still based on those assumptions so i can't be sure.
default is 0.3

NAVras-Z:
made a mistake at "A", forgot w at dot()

NAVras-Z:
Color'' = mix(max, C', A)
A = remap.y * K + remap.x , [0.99, 1.17]
K = 1 - (1-2*avg)^2, K belongs to normal range [0,1]
avg = (R'+G'+B')/3, also [0,1]
Reminds me of how it controls the colors of loading screen, a multiplier and an addition, looks similar to this one's. 
Also i can't find a meaning without joining them together,
like A = 1 + 0.18(k - 0.01/0.18) looks more like a proper weight. (A>1 saturate)
(Perhaps uf_remap here are just calculated value without real meaning? @alexkiri )
So if want to control saturation just scale A with a factor, which is the same without changing where it saturates/unsaturates.

PR #113, opened 2017-10-18
Execute build.sh during appveyor build phase using Cygwin

rajkosto:
Once you merge this, there's no turning back. Make sure you only have working templates in your build.sh because anything from the repository will be overwritten by their outputs before being placed in the zip file by AppVeyor.

After you're sure that the templates are building correctly and are equivalent to the files in the repo, you can even delete the ones in the repo, to unclutter commits.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,5 @@
 # Appveyor configuration file for bash build of packs
+version: v2-{build}
 environment:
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -1,17 +1,52 @@
-version: v2-{build}
```
slashiee: what about keeping the version formatting the same?

PR #121, opened 2017-10-19
Added Github Releases to host the artifacts

Crementif:


slashiee:
How did you set up the secure auth token?

build.sh
```diff
@@ -57,7 +63,7 @@ std_respack () {
 	params=( "$@" )
 	rest=( "${params[@]:1}" )
 	
-	#echo "std_respack $gameName"
+	echo -e "${GREEN}[Building] ${BCYAN}$gameName ${NC}to ${params[@]:1}"
```
rajkosto: echo -e "${GREEN}[Building] ${BCYAN}$gameName ${NC}for ${rest[@]}"

build.sh
```diff
@@ -1,39 +1,45 @@
...
-		echo "Making output directory $dstDir"
-		mkdir -p "$dstDir"
-	else
-		rm "$dstDir"/*.txt
-	fi
+        for f in "$srcDir"/*.txt
+        do
+                filename=`basename $f`
+                outName="$dstDir/$filename"
+                #echo "Building $f to $outName with params ${rest[@]}"
```
rajkosto: echo -e "${GREEN}[PHP] ${NC}Exec $f to $outName with params ${rest[@]}"

the php invocations are the main build step, why silence those

Crementif:
(It closed due to the reset on my fork FYI)

I changed all of things mentioned by Rajkosto in https://github.com/Crementif/cemu_graphic_packs/commit/348d136ef2276e404c21014451f151feaa1b115d which added:
* Added github releases with correct whitespace (in all of the files, blame Unix).
* Added comments on the deployment.
* Added suggested changes from Rajkosto.

I do think there should be some shortened version of it as I rather have it all not wrap to the next line.

Crementif:
Updated it with a proper description which looks something like 
image.
All links are relative so it shouldn't be an issue on either. I'm waiting for Rajkosto to accept my changes and then I think it is good to go.

rajkosto:
is +    auth_token:
+      secure: 4JvDhDJbAqoItEgXYLZ1QAX3aW5MF7d/klgYFiyCtgGJys2+C0Etd7CZwJzZ8tch
a dummy token ?

because private information like that should be set in the appveyor project settings page, not in the yml file.

Crementif:
Not a dummy token. It's encrypted ([a feature](https://www.appveyor.com/docs/build-configuration/#secure-variables) that's meant for this) so that it only works for my account. When @slashiee changes the environment variable I suggest setting it in the UI (but you should still encrypt it!). The environment variables set [should be merged](https://www.appveyor.com/docs/build-configuration/#appveyoryml-and-ui-coexistence) into the Appveyor.yml.

slashiee:
Will it just update the same release over and over again?

PR #124, opened 2017-10-24
[BotW] shadow blur removal

NAVras-Z:
@JoelAlone 
Credit @alexkiri found this shader (and also looked into it).
But yea this is a dirty workaround... can't get the same blur atm anyway so might just remove it xD
(also because the coordinates calculation looks complicated...)

Mainly for 4k and up where the 2x2 blur doesn't look like blur anymore
from alucard0712_rus:
point
from Joel:

The side-effect is that the shadow now look more aliased without blur.


NAVras-Z:
nice~

PR #128, opened 2017-11-01
[BotW] FXAA3.11

NAVras-Z:
Replace game's post-aa with pc version fxaa, more performance heavy.
FXAA was modified to take two textures, one color one luminance texture(in R).
Also format overwrite R8 to be R16 (better aa accuracy?).
AA in inventory was fxaa, dunno what in game AA was, looks like a pretty simple one.

Subpix:
Choose the amount of sub-pixel aliasing removal.
This can affect sharpness.
   1.00 - upper limit (softer)
   0.75 - default amount of filtering
   0.50 - lower limit (sharper, less sub-pixel aliasing removal)
   0.25 - almost off
   0.00 - completely off

 EdgeThreshold:
 The minimum amount of local contrast required to apply algorithm.
   0.333 - too little (faster)
   0.250 - low quality
   0.166 - default
   0.125 - high quality 
   0.063 - overkill (slower)

 EdgeThresholdMin:
 Trims the algorithm from processing darks.
   0.0833 - upper limit (default, the start of visible unfiltered edges)
   0.0625 - high quality (faster)
   0.0312 - visible limit (slower)

 Credit: Kiri, NAVras


NAVras-Z:
Screenshot:
original:
org
fxaa:
fxaa
original:
org2
fxaa:
fxaa2
However i think original in-game AA is not filtering correctly, thus the result, but I can only see that the vertex shader for luminance got unscaled resolution.

NAVras-Z:
updated in-game aa luma, but dunno if it's completely fixed since i don't understand the way it works.
Also, for fxaa, the two shaders that calculates luminance are also modified to be non-linear as instructed in fxaa doc, it should be fine since nobody else is using that shader/texture afterwards.

slashiee:
>Replace game's post-aa with pc version fxaa, more performance heavy.

that wording makes it sounds like botw has a pc version lmao

alexkiri:
Ah, excellent! So the shader actually uses the luma texture from textureUnitPS1 instead of calculating it in the shader. This appears to just work beautifully.
One question though. What does 141f4 fix? I'm curious to see it in game.

alexkiri:
how cool! both the luminance shaders start with 58b841. What were the odds?

NAVras-Z:
@alexkiri 141f pass the shifted coordinates to 58b84, they're used to calculate in-game aa's 0.5*(R+G+B)

PR #138, opened 2017-12-01
Added Temporary Hyrule Warriors 30 FPS Lock.

MelonSpeedruns:
Hyrule Warriors runs the menus at 60fps. Because of that, they're twice as fast. Locking the FPS to 30 allows the menus to be at normal speed again. Also fixed the high-resolution pack. It was causing the background of the menus to be all black.


slashiee:
Why remove everything but TV in rules.txt?

MelonSpeedruns:
Like I said, it was causing the background menus to be black. I'll make an update with only that problem removed.

slashiee:
Are all of those resolutions causing the issue?

MelonSpeedruns:
See, that, I cannot confirm. I'll check which ones are causing issues today.

PR #155, opened 2017-12-31
Created new BOTW Sharpening-only preset for Clarity

fabiograssiotto:
Hi,

For those people that want to keep BOTW original look, I created this pack that only applies sharpening.
It is good to offset the game AA. It is of course based off Clarity.
Do you think this could be added to as a new graphics pack?

Thanks,


slashiee:
what's the difference between using this pack and just modifying clarity's settings to just use luma sharpening?

fabiograssiotto:
It is exactly the same, only for people that do not want to or understand how to change #define statements. Perhaps it could be made into a profile for Clarity?

mrtawsiti3:
i like this PR alot

slashiee:
Yeah, a profile for Clarity makes more sense.

fabiograssiotto:
Changed this into a Clarity Preset on the Clarity Presets folder.

PR #156, opened 2018-01-02
Rename bd8bba59e2149449_00000000000003c9_ps.txt to 37040a485a29d54e_0…

tet666:
Just an update for contrasty so it works in botw 1.4.0+


tet666:
Just an update for botw 1.4.0+

PR #160, opened 2018-01-06
Fixed NoMMTriforce on Nvidia

Darksim150:
Unknown if the fix breaks something on AMD.


Crementif:
Is this issue related to Nvidia's texture corruption? If so, the format change shouldn't prevent it from being affected, from my experience.

Crementif:
Checked by Jamie, it doesn't look like there was any corruption. Thanks a lot for the fix!

PR #172, opened 2018-01-25
[HW] Typos in the normal resolution file

nosklo:
Fix some errors in the normal resolution file, that are affecting the ultrawide modes. 
The original file uses scaleFactorY for width which shouldn't matter for normal scaling but makes some different values when using non-16:9 proportion.


slashiee:
they actually did matter? huh, the more you know then.

PR #175, opened 2018-01-31
Created Wind Waker HD FPS Slowdown Fix Pack!

MelonSpeedruns:


Magmarock:
Hi, Could I please get a little more detail as what this does exactly. How does it reduce frame loss and will it be needed in further Cemu updates

slashiee:
I believe it disables network connectivity stuff from last I remember.

Magmarock:
Thanks for getting back to me. I'd like to request if I may that little descriptions for FPS and other non resolution related graphics packs gets included. Having a higher FPS is always nice but it's impotent to know how it's being done encase in works by disabling a feature you want or something.

PR #179, opened 2018-02-07
Static FPS mods from https://forum.cemu.info/showthread.php/421

epigramx:
After discussions I had with Slash and NoskLo, this is a short list of mods (only 30/45/60) from https://forum.cemu.info/showthread.php/421. I'd prefer it if they had a "see readme.txt" in the rules.txt name but it's probably too spammy for this repo so I followed naming convention; oh well, if users want more info to make their own fps target or get another version, they can just visit the forum thread.

PS. Be sure to read the thread or the readme of the packs if you don't understand what they are. The gist is that it's a simplified version of fps++ to fix graphics judder while it also includes the rest optimizations included in it; the caveat is it drops dynamic fps target support so the user must be able to attain said target.

### DIFF
--- a/Modifications/BreathOfTheWid_static_30FPS_mod/patches.txt
+++ b/Modifications/BreathOfTheWid_static_30FPS_mod/patches.txt
@@ -1,0 +1,108 @@
+[BotwFPSV208] #i.e. v1.5.0
+moduleMatches = 0x6267BFD0
+
+#Cave
+codeCaveSize = 0x1C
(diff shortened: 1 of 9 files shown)


epigramx:
By the way, a good point passed on to me is that a display of 120Hz might want to use 40FPS instead of 45 and 48 instead of 45 on 144Hz and that there could be a php generation of more packs. However, I got the general impression nobody wants a spam of many packs in the default release and power-users should probably read the readme and make their own, if they don't download one from the thread.

PR #189, opened 2018-02-12
Added Play-as mods for Hyrule Warriors! (Both Beast Ganon and Giant Cucco)

MelonSpeedruns:


slashiee:
can't rebase because of conflicts, lmao

squash'll do

PR #209, opened 2018-04-05
Splatoon Ultrawide fixes

Crementif:
See https://github.com/slashiee/cemu_graphic_packs/issues/207#issuecomment-379059749 for all the fixes


slashiee:
might be better to put some of these into a separate pack? there's too many bugs on the online side of things

Crementif:
Are the bugs any worse then the bugs that are there? Would be ok with creating a separate pack tho. Don't know of any bugs that are there on 21:9, although I did realize a minor mistake in 1 shader in particular.

slashiee:
Should be pushed to a separate pack, I can accept it then

Crementif:
I've added them as a separate pack. The commit and file history wasn't particularly clean either, but I've just included all the ultrawide related fixes. Having a separate pack is still questionable, but I think these ultrawide projects are just too much to maintain/begin with. I'm probably not bothering too much with future endeavours. Scraping the branch is also an option... :sweat:

slashiee:
This should be fine...

Issue #143, opened 2017-12-07
When scaling the shadow map resolution in rules.txt, the blur part (penumbra) needs to be scaled to the same level in the shader


getdls:
When scaling the shadow map resolution in rules.txt, the blur part (penumbra) needs to be scaled to the same level in the shader.

![4kscaled](https://user-images.githubusercontent.com/25286924/33717745-62bc3cc0-db5b-11e7-97c2-1a4d7e351336.png)

Otherwise you end up with a shadow like below. I assume we have the same problem for other games.
![4kunscaled](https://user-images.githubusercontent.com/25286924/33717794-7ee08c6c-db5b-11e7-9ea8-abab14e38a7b.png)

The problem is that this scaling follows the shadow map, ie high res shadow pack not the main rules. @slashiee , any suggestions on how you want to organize this?

- Stick all of the shadow shaders and scaling in the source folder 
- Generate separate shadow resolution folders by adding build logic
- Leave as is and hope ppl read the header and scale shadow shader to the same multiple as high res shadow rules
- Other suggestions?

slashiee:
Probably put the higher resolution shadows in the packs themselves? Dunno how other people would think of that though

Crementif:
We already scale the blur in other blur and shadow shaders, so I'd go with that. Something like this https://github.com/slashiee/cemu_graphic_packs/commit/47264be1b8836de48aaca79b925bb6db28d17fbc would be sufficient, so some code like
```<?php 
include 'Source/functions.php'; 
$fullWidth = $argv[1]; 
$fullHeight = $argv[2]; 
$scaleFactorX = always_decimal_format($fullWidth / 1280.0); 
$scaleFactorY = always_decimal_format($fullHeight / 720.0); 
?>
#version 420
#extension GL_ARB_texture_gather : enable
#extension GL_ARB_shading_language_packing : enable
// shader 1f83c0d47b1c4c34
// Used for: Background Blur
const float resXScale = <?=$scaleFactorY?>; 
const float resYScale = <?=$scaleFactorY?>; 

~Rest of shader code~
```. Having people change stuff hasn't been working wonderfully in the past.

getdls:
That's what I'd vote for as well.

Maybe round to 1 <?=round($scaleFactorX)?> so we don't downscale shadows. 
hmm.. all of the other scaling is probably off below 1280x720 now that I think about it :(

getdls:
Alright, scale in source folder it is then.

[...]

Issue #149, opened 2017-12-18
[Splatoon] High resolution shadows makes models look blocky


slashiee:
![image](https://user-images.githubusercontent.com/15317421/34089675-dd662e5a-e366-11e7-84e5-197000f6af13.png)
![image](https://user-images.githubusercontent.com/15317421/34089683-eebf3c50-e366-11e7-9c94-2e4f1861b188.png)

Crementif:
Might be a problem with reflections, which are really low res. It needs some mip scaling (256x256), but I'm not quite sure if I know how to do that without breaking stuff.

Crementif:
Can't find a way to replicate the bug, it looks fine wherever I checked. Does this only happen on the match end screen?

slashiee:
It also appears on the end screen for any single player map.
![](https://i.redd.it/tea8sb9cxqlz.png)

Crementif:
The texture that might be causing an issue is a 4x4 texture, so you might want to upscale (or remove) that. Editing the shaders didn't do anything for me for some reason.

Crementif:
Fixed the issue with commit https://github.com/slashiee/cemu_graphic_packs/commit/91bdffd0dae455479c738c5a8e8568d31c4450ce supposedly.

slashiee:
Might be better to put it into main pack like what we agreed with in #143

Crementif:
I don't think it has something to do with the blurring at the edges, it's probably just not scaling properly. If I had to guess it's still a 4x4 texture that needs scaling, but that's way to common to attempt. I don't know why he only tried scaling this shader, it's only used on the end-screen? I had to replay the story mission to dump this shader which only was there when I reached the end-screen.

Anyway, I'll put it in there, there's probably no harm in scaling those shaders.

PR #210, opened 2018-04-07
Move shadow packs to main pack

slashiee:
As talked about in #143


slashiee:
@Crementif The shader edit that you made for the high res pack is also in the main pack... Does that shader fix the issues from #149?

PR #218, opened 2018-05-13
Switch to Travis CI with multiple builds

Crementif:
This will swap over to Travis which offers native linux builds. It's, against our expectations, slower or equal (probably due to lower specs), then our current Appveyor build. It does allow for a cleaner build and also implements a way to add random resolutions people would like to be added in our new 2 parallel building method. It builds Github Releases that contain multiple assets. This'll also make the common downloaded resolutions be more on their own so that people won't have loading problems. Also, the building process contains less initial spam due to the Travis' line folding. If there's an error from the php script it'll unfold automatically so no information is really hidden.


Crementif:
Only thing left for you to do is to change the https://github.com/Crementif/cemu_graphic_packs/blob/master/.travis.yml#L26 line to a secret (the one you send isn't correct, I needed the token from https://github.com/settings/tokens with all the repo options toggled). Plus, removing Appveyor webhook in this repository settings (from what I recall) and probably some other appveyor stuff. Here's what you'll need to input 
image. I'll be unavailable next week so it's probably better to wait until then before merging. There's also some other changes that'll have to be debugged and some other website changes.

slashiee:
It should be worth noting that these changes will break @nosklo's updater application.

Crementif:
Yup, unless he made some precautions. Nosklo could probably have some till
until then to update the hook using one example release that is already up
on https://github.com/Crementif/cemu_graphic_packs/releases.

slashiee:
Apologies for this, but I have to merge it now because a lot of changes got reverted when you removed the resolutions.

PR #223, opened 2018-05-28
Update SplatFX

slashiee:
Updated to use current Clarity shader (presets built-in, new options). Current issue is that the main shader is getting overly saturated? Need someone to look into it.


Crementif:
I'll look into it later. I'd also preferably make it use a different render step as having two shaders is just plain inconvenient.

slashiee:
Before you commit your new version, can you grab the latest presets from 91e2937ee363042b5527457e3e29eccf1ab3886e please? @Crementif

slashiee:
To clarify:
- 4a377a50eda82af8 has the over-saturation issue
- f502868a515bbfb4 is (at least) used for Flounder Heights, it does not have the over-saturation issue

slashiee:
Rebased to current master.

PR #228, opened 2018-06-26
Add 768p and 1536p to uncommon resolutions pack

Morph1984:
1366x768, commonly known as 768p or "laptop" resolution is found on some (cheaper) laptops, monitors, and TVs.
2x resolution of 768p is 1536p. Available for downsampling perfectly to 768p for better image quality if there is enough GPU speed/for games that do not depend much on GPU power.

720p performs marginally better than 768p but delivers worse quality (due to imperfect upscaling) on 768p screens.
For 900p, it performs worse than 768p and also delivers worse quality (due to imperfect downscaling).


build.sh
```diff
@@ -57,6 +57,7 @@ res_540p=( "Performance" 960 540 )
 res_720p219=( "Performance" 1706 720 )
 res_720p489=( "Performance" 3840 720 )
 res_720p=( "Performance" 1280 720 )
+res_768p=( "Performance" 1366 768 )
```
slashiee: I'd probably put 768p under Quality since it's higher than 720p, which is what most Wii U games run at.

PR #232, opened 2018-10-08
Rename all Xenoblade mods from lasyan3 for better sorting and underst…

lasyan3:
Hi, it's me again :)
Right after the last pull request I realized my mods naming is not convenient, so I decided to rename them al for a better understanding of what they do and a better sorting in Cemu. I'm pretty happy with the result.
Besides, some mods were not properly tested and had issues, fixed now.


Crementif:
oh, sorry, just seeing this now. I'm not completely sure how to handle this case. For the folders I would probably prefer something like `XenobladeX_CheatsGPL_MorePropeIncome` so that the cheat packs would all be clustered together instead of them being split up by random other xenoblade packs. The same goes for the names inside the graphic pack window, something like `Xenoblade Chronicles X - Cheats - GPL - More Prope Income` (I also think the names probably should be the same as the folder names since people probably would want to edit these packs and wouldn't be able to find them) so that they too would be clustered. Lemme know what you think, I'm just giving some thoughts. I'll merge it whenever you think it's best.

lasyan3:
Hey Crementif,
I totally agree with you, but I won't use "cheatsGPL" as I don't consider all my mods as cheats, and not all of them concern gameplay (there are some for the music, for the graphics, etc.). I think I'll rename all with prefix "mod" and remove spaces.
Example: with the mod "Xenoblade Chronicles X - Collectibles - Catch range mod" will give "XenobladeX_mod_CollectiblesCatchRange". This way all folders should be grouped together but will keep a naming similar to the one displayed in Cemu.
In any case, all of that is hopefully temporary, as I really hope rajkosto will release a cemuhook version with support of the version 3 of graphic packs. This way I'll be able to add descriptions and presets for the mods, which will be awesome!
Until then, I'm going to rename all the mods and pull the changes (I don't know yet if it's possible to edit this pull request, I'll see).

lasyan3:
Commit done! What do you think?

Crementif:
Oh, yeah, "mod" is totally fine too! I also hope that we'll get native patches (though I also know that they'll need rewriting) soon or Cemuhook support soon. Thanks for making the changes!

PR #251, opened 2018-11-04
HyruleWarrior crash fixes updated for 1.14

klittle81:
Add HyruleWarrior fixes updated for 1.14


klittle81:
just adding the HyruleWarriors crash fixes that I updated to work with cemu 1.14

PR #277, opened 2018-11-20
Fixed 8x8 Texture redefine in Bayonetta 1 Res pack

Zalnor:
Changed the format section of the texture redefine from "formatsExcluded =" to "formats = 0x001a" allowing the game to boot again with the res pack enabled.

As a side note I did not see any other texture formats using 8x8 resolution besides 0x001a

### DIFF
--- a/Source/Bayonetta_Resolution/rules.txt
+++ b/Source/Bayonetta_Resolution/rules.txt
@@ -59,7 +59,7 @@ $gameWidth = 1280
 $gameHeight = 720
 
 [Preset]
-name = 3840x2160
+name = 3840x2160 4K
 $width = 3840
 $height = 2160
 $gameWidth = 1280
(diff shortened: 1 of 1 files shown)


Source/Bayonetta_Resolution/rules.txt
```diff
@@ -1,168 +1,168 @@
...
+
+[Preset]
+name = 3200x1800
+$width = 3200
+$height = 1800
+$gameWidth = 1280
+$gameHeight = 720
+
+[Preset]
+name = 3840x2160 4K
```
Crementif: makes stuff a little inconsistent...

Crementif: Oh, I just meant to make a comment out of it, not a review. Ah well. I don't seem to be able to edit the pull request, but it's fine to keep it as is.

Crementif:
@getdls Seems like Cemu doesn't like the fact that you had an empty thing redefine when it tries to parse the expression. You probably should patch your automatic program to not do that kind of stuff, just want you to be aware of this issue. Just want you to be aware of it.

getdls:
Thnx for the heads up

PR #282, opened 2018-11-26
fixing some stuff

Morph1984:
can I use 2 decimal places instead of 1 for multiplier, see commit changes.

Also, why is $gameWidth and $gameHeight for resolutions_1080.md 1280 and 720 respectively?

Finally, removed 4300x1800 in botw rules.txt (consistency)

### DIFF
--- a/Source/AmiiboFestival_Resolution/rules.txt
+++ b/Source/AmiiboFestival_Resolution/rules.txt
@@ -1,0 +1,129 @@
+[Definition]
+titleIds = 00050000101C6400,00050000101C6500,0005000010190100
+name = Resolution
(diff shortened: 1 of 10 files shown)


Morph1984:
done
some others were not touched yet (those with $dither or other things attached

skalfate:
also morph, replace the // symbols in the 16:10 resolutions with a # to make them work properly. I had to fix up my commit just a few minutes ago. People reported the resolutions disappeared.

skalfate:
Also if you plan to redo the single decimal rounded resolutions I did into 2 decimal numbers resolutions, make sure to try them out and see if they work properly using stretch. 

The reason I say that is because the 900 resolution was a 2 decimal ratio 1.25, and it had the black bar at the bottom.

PR #323, opened 2019-01-13
[BotW] add No MM Triforce mod

delta592:
Converted the existing v2 rules.txt into v3 using exact TextureRedefine values. Not sure if anything under the hood changed that would affect the values. Tested this by playing in Master Mode. There were no apparent visual or performance issues.


slashiee:
Would rather prefer a shader edit but I guess this is fine for now

PR #326, opened 2019-01-21
faster fps++ (use better fence)

Morph1984:
in lieu of @epigramx changes to static fps mod which uses better fence instead of best fence.
https://github.com/slashiee/cemu_graphic_packs/commit/a85380ff8b39b2c396c9638729fb77a420213d80

### DIFF
--- a/Mods/BreathOfTheWild_FPS++/patches.txt
+++ b/Mods/BreathOfTheWild_FPS++/patches.txt
@@ -126,8 +126,8 @@ _arrowTimeDrain = 0x00000114
 
 
 #break all forms of frame limiting
-0x031FAAFC = bla _fenceNeg1 ;Best fence (make sure GPU is never more than 1 frame ahead)
-;0x031FAAE8 = li r0, 1 ;Better fence (make sure GPU is never more than 1 second? ahead)
+;0x031FAAFC = bla _fenceNeg1 ;Best fence (make sure GPU is never more than 1 frame ahead)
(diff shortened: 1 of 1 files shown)


Xalphenos:
Better fence is better for some people.  Not all.  I talk about why [here.](https://github.com/slashiee/cemu_graphic_packs/issues/308#issuecomment-450078094) The general consensus is that best fence prevents milk water.  Since the current goal of fps++ is less bugs we would need a lot of testing to make sure milk water wasn't more likely to happen with better fence.


Now for anyone who's pc is good enough that botw selected the 30 fps mode then there is no difference between best fence and better fence outside the first 500 frames that best fence does a full fence on.  After that best fence does (framecounter + fence value) - 1 where fence value is always 2 if in 30 fps mode.  Best fence overwrites fence value with 1 regardless of what it was.  So assuming pc is good enough, and we're past the full sync frames.
Best fence:
wait for framecounter to equal (framecounter + 2) -1.

Better fence:
wait for framecounter to equal Framecounter + 1

Those are the same thing.  Just to clear up some confusion on the two.  But best fence did a full fence for the first 500 frames to prevent milk water.  The problem is when the pc isn't good enough and botw selects 20 fps mode resulting in the fence value being 3.  In my testing I've never seen the fence value be anything other than 2 or 3.  But best fence would take into account other variables that may be needed where as better would not.

That said if there was enough testing done that proved there was no difference in best, better, skip I would have no issues changing it.

Morph1984:
closing first

PR #358, opened 2019-05-01
[BotW] Too Many Options pack

Crementif:


Crementif:
@Alstura 

Okay, so I think I'll review/give my opinions on this. First of all, this does increase the size of the zipped repository from 1.45MB to 2.37MB zip size which is pretty large for 1 mod. Hence why I think that it should be reduced. I suggest you minimize this mod's footprint by removing presets that don't function or look horrible. The rules.txt files take like 2/3 of all of the space and I doubt that most of these presets look usable. It should be somewhat selective, IMO. Clarity consists of the only the best presets, while it doesn't offer 200 sliders for you via separate packs. I think making presets like that would be limiting however, but I think that's definitely more in the spirit of the graphic packs repository.

Second thing, these are all mods while you've put them into Enhancements, which some of these definitely are not. And, I would like these to be all one sub folder inside that folder.

Third thing, I don't like a separate credits file. If you do that, it's not part of the whole "graphic packs" repository, but makes it look like an separately installed mod.

Fourth thing, I'm not fond of the jokey nature of some of the presets (like the section headers), it just feels out of place if the reset of the graphic packs repository are more or less professional and try to be as clear as possible. Makes it feel more like a separately installed mod, again.

---
Maybe I'm just too strict and I'm limiting options, but there's already been complains about the "complexity" of graphic packs. I don't agree with that sentiment (I think it could use some reorganizing though), but this pack pretty much doubles the amount of dials.

I really like the work, so I think after slimming it down a bit by making the presets more selective, it would definitely be a super useful/wanted thing to have, especially since there's some awesome tweaks that would probably also fit outside of the too mini options pack very well. I'm totally not saying that this shouldn't be merged.

@slashiee could maybe give his opinion too, I'd like to hear your and slashiee's opinion on this.

slashiee:
i don't have time to look at it rn, but doing what crementif is requesting for would probably be a good start

slashiee:
shaders will need to be updated for vulkan in order to be added to the repo btw

RandomGuyJCI:
With the implementation of v5 packs and nested menus, you may want to change the structure of the files.

slashiee:
I'm gonna close this since major changes are gonna be pushed to the repository soon. Feel free to create a new PR once you get TMO updated.

PR #365, opened 2019-05-17
[BotW] 38" ultrawide resolution (3840x1600) added

arkste:
i've added the resolution 3840x1600 (21:10) for BotW.
i own a LG 38WK95C-W which runs at 3840x1600 and its resolution was missing.


Crementif:
Well, I'm assuming this works, so I'm merging this.

PR #371, opened 2019-05-30
Remove duplicate path folder

RandomGuyJCI:
bruh there were 2 separate folders


Crementif:
Fixed by a different PR that also fixed this issue and also made it clear that it's for Intel GPU's only since we've had a few people that tried using it with an Intel CPU.

PR #372, opened 2019-05-30
Rename the Intel iGPU path

Morph1984:
Removes duplicate path and renamed for clarity


Morph1984:
This renames "Intel iGPU" and "Intel integrated GPU" into "Intel Integrated GPU (Only for Intel GPUs)"

PR #378, opened 2019-06-10
Added NES/SNES/N64 Undarken packs, and added more SNES Title ID's to its Resolution pack

Zalnor:
NES/SNES Values for how much brighter are not exact as I don't have another emulator to compare them to currently.


slashiee:
From the pics you posted, SNES seemed fine at normal brightness. I'd check to make sure NES is accurate though

PR #383, opened 2019-07-07
[BotW] Circular blur: cut costs with linear

NAVras-Z:
By sampling at the 2x2 center of 4 adjacent texels, we get the average of the squad at once. Reducing 64 samples to 16 at 1440p.
I believe this is what the game was doing as well:
1. the result texture is set to half the original res.
2. the shift in coordinates is 1 seen in the vertex shader.

It means that instead of sampling at the texel center of textureUnitPS0, it samples at the 4 corner of the 2x2 square. For GPU's linear interpolation, each sampling gives the average value of 4 neighboring texels, together it gets all 16 texels in the 4x4 square... with just 4 fetches.

We can mimic this in our circular blur shader. The result at 1440p is slightly less blurry than native 720p when using sheikah scope. But I don't think it's completely because of this shader, as the sheikah scope uses this in combine with a fixed radius blur shader, so less blur is expected.

### DIFF
--- a/Resolutions/BreathOfTheWild_Resolution/cb0e6e8cbec4502a_0000000000000079_ps.txt
+++ b/Resolutions/BreathOfTheWild_Resolution/cb0e6e8cbec4502a_0000000000000079_ps.txt
@@ -1,29 +1,29 @@
 #version 420
 #extension GL_ARB_texture_gather : enable
-#extension GL_ARB_arrays_of_arrays : enable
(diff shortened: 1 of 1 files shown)


Crementif:
This sounds pretty awesome, especially since I think this shader is also used in situations besides the camera, scope and battle for the far distance if I remember my debugging session correctly and it being expensive.

PR #385, opened 2019-07-25
[BotW] Fix 960x540 background ambient light, bloom

NAVras-Z:
...And other texture cache's mismatch warnings.

Uncomment 128x64 for a scope DOF fix at line 760 if HUD setting is Pro, since the game doesn't draw the sound HUD.

Resizing 0x1,0x1a has the risk of breaking textures, tho there wasn't any game texture matching the resolution and format in Nsight resource tab. I haven't notice any obvious breaking while playing.
Separate slices of Cubemap being caught by 32x16 resize is not fixed. It needs too many 0x816, 0x1a resizes to not have texture mismatches.
Also I changed to file to Windows(CR LF), thus the diff...

What was 1/10 resolution intended for?

1/10 resized, missing bloom and ambient lighting('red filter').
old
native, the fixed version looks the same.
native

I have no idea what the game is doing with this, it's prepared with blurs but not getting used at the time of the captures. Still, I've added the rule to avoid breaking it.

### DIFF
--- a/Resolutions/BreathOfTheWild_Resolution/rules.txt
+++ b/Resolutions/BreathOfTheWild_Resolution/rules.txt
@@ -312,7 +312,7 @@ $heightfix = 0
 width = 1280
 height = 720
 formats = 0x001,0x005,0x007,0x019,0x01a,0x41a,0x80e,0x806,0x816,0x820
-formatsExcluded = 0x008 # Game Load Opening Background Image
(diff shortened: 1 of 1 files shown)


NAVras-Z:
Didn't get any extra warning besides the Cubemap one while roaming around Hyrule. Should be stable, need to push this for another commit.

PR #388, opened 2019-07-28
[BotW] Resize letters box to make text less blurry

NAVras-Z:
For many cutscenes the game draw the subtitles on a 500x94 box first and apply the box with the letters onto the screen. Resized the box to make subtitles less blurry.
Text bloom/effect use another 560x150 box, it's good not having to resize it.
orig
up
The pop up message is also sharper after resizing, but looks like it might have been over-sampled a bit.
orig2
up2


NAVras-Z:
Trying to resize Hyrule Compendium text boxes, it's a mess, rule for this one's box catches catches another text's bloom, resizing the bloom then further catches something else...
[rules.txt](https://github.com/slashiee/cemu_graphic_packs/files/3482865/rules.txt)
No resize
no
Upscaled. text, ancient gear and giant ancient core icons
screenshot_2
the icons use 96x96 texture while being 58x58

PR #391, opened 2019-08-01
Add scaling filters

NAVras-Z:
Add your titleID in rules.txt for them to work.
Updated Cemuhook's Bicubic and Lanczos scaling shader.

Bicubic was recently optimized by the author at OBS to use 9 taps instead of 16. Cemu's Hermite is the Catmull-rom preset.
Fixed Jinc (Lanczos) shader's divide by zero check. So there is no longer black dotted lines artifact when not zooming.

### DIFF
--- a/Filters/Bicubic (4 taps)/output.glsl
+++ b/Filters/Bicubic (4 taps)/output.glsl
@@ -1,0 +1,50 @@
+
+#version 420
+
+in vec2 passUV;
+layout(binding=0) uniform sampler2D textureSrc;
+uniform vec2 textureSrcResolution;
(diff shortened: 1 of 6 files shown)


NAVras-Z:
Fixed Cemu's Bicubic
When put B-spline's b&c and [|d−1|=f+1; |d0|=f; |d+1|=1–f; |d+2|=2–f](https://vec3.ca/bicubic-filtering-in-fewer-taps/) into the [Cubic weight function](http://avisynth.nl/index.php/Resampling#Bicubic_resampler), everything in w() matches up. Except the shader is not taking distance as weight, nor does it uses round down for coordinates as supposed to.

PR #394, opened 2019-08-06
[Wind Waker HD] Added Picto-Box workaround

Vladimir-1799:
Added workaround for 'Lenzo's Research Assistant' quest. Lezlo will now accept any picture shown. Works with USA and EUR versions.


Crementif:
Oh, this is awesome! What did you meant with the [readme](https://github.com/slashiee/cemu_graphic_packs/blob/master/Workarounds/WindWakerHD_PictoBox/readme.txt) though? Loadline shouldn't be any different then the wud copies, and there aren't any updates. Just curious.

PR #400, opened 2019-11-07
Add more lod bias granularity, as the current high preset is too much

Esmaperdut:
Add more lod bias granularity, as the current high preset is too high and produces lots of shimmering even at 4k. -1 or -2 are the presets recommended over the internet as they add detail without producing too much shimmering.

### DIFF
--- a/Enhancements/BreathOfTheWild_LODBias/rules.txt
+++ b/Enhancements/BreathOfTheWild_LODBias/rules.txt
@@ -10,27 +10,39 @@ name = Normal (Default)
 $lodBias = 0
 
 [Preset]
-name = Lowest
+name = Lowest (+16)
 $lodBias = +16
 
 [Preset]
-name = Very Low
+name = Very Low (+8)
 $lodBias = +8
 
 [Preset]
-name = Low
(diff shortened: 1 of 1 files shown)


Crementif:
Okay, I made my own fixes to the LOD pack. Thanks for contributing the idea of having the number there (will be helpful for people knowing the technical details) and indeed for the idea of having more helpful presets added, since for play-ability the only slightly lower ones are really helpful.

However, I had a few problems with the pull request itself, so I thought it'd be better if I spend some time doing some proper work. The Normal+, Normal++ and Normal+++ names wouldn't be great and the numbers don't really mean anything.

I implemented a proper description that was previously missing (tried to explain it in a simple way), the names of the preset also tells you what "Low" is, since it's now "Lower texture detail", tried to make a literal range of levels (Highest->Extremely High->Very High->Much Higher->Higher->Slightly Higher) and made the presets symmetrical since you only made presets that add granularity for lowering the texture of detail.

Esmaperdut:
Perfect, I like the new naming solution you found.

Thanks!

Joan,

PR #409, opened 2019-12-21
Merged Workaround Packs

Zalnor:
I merged the workaround packs for BOTW into GPU catagories for OpenGL and updated their version to 4.
Along with this removed the workarounds for intel flipped lights and reduced bloom as I tested and concluded that its a non issue on latest drivers for iGPUs
I also updated LWZX Crash pack to V4 as well.


slashiee:
#411 reorganizes workaround packs as well, personally more of a fan of how it's done there.

PR #410, opened 2019-12-25
[BOTW] add rendererFilter to LWZX Crash Pack

shinrax2:
add "(OpenGL)" to title and rendererFilter to this pack since it creates crashes on vulkan and is not nessecary due enforcement of GX2DrawDone on vulkan


slashiee:
i think it's better to just force it to opengl renderer without mentioning it's only for that

shinrax2:
ok ive updated it accordingly

PR #411, opened 2019-12-26
Update BotW packs for Vulkan

Crementif:
But now done properly!

Basically, a bunch of improvements were made to the script. The previous attempt at this conversion was quickly followed by a rollback since I realized that the script was overlooking certain things that made most of the packs hit or miss whether they would work. A few things missing were:
- It only tested the values from 1 preset. Now, each shader gets compiled per each preset, like what Cemu would do. It also merges the changes done for each preset into one. This should solve cases where one shader would define things separately or repeatedly from preset to preset. 
- All* of the shaders are tested to see if they use the converter used the right values for the locations for Vulkan.

Both of these *should* mean that they should both compile and be linkable in Vulkan, which means that I don't have to test each individual shader to see if they work. I will release the two scripts (one used for converting, one used for checking the right values for the locations) tomorrow so that other people might be able to help, if they want. It's fairly straightforward now at least.

*There were some shaders that I just wasn't able to dump with Cemu 1.16.0 with my 8000 shader cache (and only one that I actually had to manually fix!) (I just enabled shader dumping and then loaded the game so that it dumps everything while the cache is compiling). I think these are just odd-ball shaders that are likely converted correctly, but just aren't verified. Or just too outdated. Here's the list though:
```
\Enhancements\BreathOfTheWild_Shadows\ffe0e8c84f6e8da9_000003c000009269_ps.txt
\Resolutions\BreathOfTheWild_Resolution\01ba1a725afa0b96_0000000000000000_vs.txt
\Resolutions\BreathOfTheWild_Resolution\8cab2ed476b991ea_0000000000000000_vs.txt
\Workarounds\BreathOfTheWild_KakarikoShadows\8e9e804a3cd384c3_00000000000ff259_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\2a045e62b99640b5_00000001e1e21c49_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\5b04d28f319a1d6d_0000000f0f0ff249_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\6240983542445009_00000001e1e21c49_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\70a79265ab7d498a_0000000f0f10e249_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\751393d6cbcd0c76_000003c3c3fc9249_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\8e9e804a3cd384c3_00000000000ff259_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\99122c964b6a1a19_000000000001fe49_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\9c9c75d87bb9b4ef_0000000f0f10e249_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\b2cf5924b04ea6a4_0000000f0f0ff249_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\b2e1c9ce2f160aab_000000003c3c4389_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\b5b53042daaad65c_00000001e1e21c49_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\bf72a1fd35401be2_0000079749255249_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\cd1eba6802df67f7_0000000f0f0ff249_ps.txt
\Mods\BreathOfTheWild_GlowLightsColors\fb135efcb4540fb1_00000078787f9249_ps.txt
```


Crementif:
Also, the shaders contain some verbose output. It's not really necessary for anyone really, since the output is mostly related to Vulkan locations I mentioned above. Which should just fully work since most of them were verified with that verify script.

slashiee:
commit fixed a file conflict

PR #417, opened 2020-01-18
Convert a lot of graphic packs to V4

Crementif:
Didn't update the docs (will do that tomorrow), but I manually checked (didn't verify them with the script but I did check if there wasn't anything too crazy that might screw things up and if there were any other signs that lead to issues) each pack to make sure that all the included ones are 99% safe.

I also didn't convert 5 graphic packs since they contained signs that needed to be manually checked or at least examined more:
- \Enhancements\TwilightPrincessHD_Bicubic (weird filter shenanigans)
- \Resolutions\DevilsThird_Resolution
- \Resolutions\TwilightPrincessHD_Resolution (this one just needs to be fully verified since it's popular enough and has like 27 shaders)
- \Resolutions\LegoStarWars_Resolution
- \Resolutions\TokyoMirage_Resolution (this one could also be verified with the script)

~~I hope I didn't make too many mistakes here.~~

### DIFF
--- a/Enhancements/Bayonetta_Contrasty/21e6bc9b0cdbe8d7_00000000000003c9_ps.txt
+++ b/Enhancements/Bayonetta_Contrasty/21e6bc9b0cdbe8d7_00000000000003c9_ps.txt
@@ -1,10 +1,36 @@
 #version 420
 #extension GL_ARB_texture_gather : enable
 #extension GL_ARB_separate_shader_objects : enable
+#ifdef VULKAN
+#define ATTR_LAYOUT(__vkSet, __location) layout(set = __vkSet, location = __location)
+#define UNIFORM_BUFFER_LAYOUT(__glLocation, __vkSet, __vkLocation) layout(set = __vkSet, binding = __vkLocation, std140)
+#define TEXTURE_LAYOUT(__glLocation, __vkSet, __vkLocation) layout(set = __vkSet, binding = __vkLocation)
+#define SET_POSITION(_v) gl_Position = _v; gl_Position.z = (gl_Position.z + gl_Position.w) / 2.0
+#define GET_FRAGCOORD() vec4(gl_FragCoord.xy*uf_fragCoordScale.xy,gl_FragCoord.zw)
+#define gl_VertexID gl_VertexIndex
+#define gl_InstanceID gl_InstanceIndex
+#else
+#define ATTR_LAYOUT(__vkSet, __location) layout(location = __location)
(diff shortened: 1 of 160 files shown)


slashiee:
I trust that these changes are fine, especially after all the work you've done to the script after the first initial commit.

@getdls Are you able to verify that Twilight Princess HD shader edits work fine with Vulkan? You'll have to convert the pack itself.

Also: Tokyo Mirage Sessions #FE has a major graphical issue in Vulkan, but the shader edits should still be checked there as well. I'd prefer to get all the packs updated to v4 before we start working on v5.

getdls:
@slashiee I can run a smoke test on a 2080 card, but that's kinda pointless considering that it's pretty much AMD and Intel users that run Vulkan.  

I figured that was that you guys/gals where doing with all that conversion and checking stuff.

getdls:
**These work with Vulkan as is**

- LOD pack 
- Distant fog 
 

**Updated for Vulkan**

- Regular res pack - Shaders updated, UI scaling removed    
- Contrasty - Shader updated 
- FancyFX is broken - removed   .w /LOD shader based effects don't work in with Vulkan (Fire, heat haze, plant drool etc). 
- Bicubic crashes - removed .glsl doesn't work with Vulkan

slashiee:
alright

getdls:
jeebus f... merge errors.. abort abort!

Crementif:
> **These work with Vulkan as is**
> 
> * LOD pack
> * Distant fog
> 
> **Updated for Vulkan**
> 
> * Regular res pack - Shaders updated, UI scaling removed
> * Contrasty - Shader updated
> * FancyFX is broken - removed   .w /LOD shader based effects don't work in with Vulkan (Fire, heat haze, plant drool etc).
> * Bicubic crashes - removed .glsl doesn't work with Vulkan

Which game are we talking about here?

Crementif:
If we're talking about TPHD here, the scaling filters don't work with Cemu at the moment so I think we can just leave it.

slashiee:
I think it's best to get rid of the TPHD Bicubic pack and wait for Cemu to add support for global packs so we can just use the Filter packs instead

Crementif:
Yup, agree.

getdls:
https://github.com/slashiee/cemu_graphic_packs/commit/55de532386ed7852fdbdd006c54e09dc2d62ef56 All done. 

If we get some multi selection option in v5 I'll move back multi-pass blur/FancyFX to the _resolution folder. Right now it doesn't know what the main settings are and will look like crap when it's wrong.

PR #429, opened 2020-04-24
[FatalFrameProjectZero]60FPS pacth update

lilystudent2016:
Support for the widely used version in chinese-using areas, translated by chinese folk and based on JPN version.It will be quite convenient for chinese-using areas because most players choose this translated version to play. Can you add it to the patchs.txt?Thank you!


lilystudent2016:
I`m so sorry that I wrote the wrong title

PR #440, opened 2020-05-12
Pictograph fix

JuanCarlosM1:
Using just that part of the code, the camera returned to working properly at resolutions lower than 1080p, on ultrawide resolutions it remains the same before this fix.

Video showing the fix:https://streamable.com/ovtvvg

### DIFF
--- a/Resolutions/WindWakerHD_Resolution/rules.txt
+++ b/Resolutions/WindWakerHD_Resolution/rules.txt
@@ -356,20 +356,21 @@ height = 33
 overwriteWidth = ($width/$gameWidth) * (60)
 overwriteHeight = ($height/$gameHeight) * (33)
 
-### pictograph. If you try to fix this, verify that saved pictures can be loaded from album.
 [TextureRedefine] #Pictograph
 width = 1600
 height = 912
 formats = 0x019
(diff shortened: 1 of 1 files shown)


slashiee:
Have you verified that the pictograph works on all resolutions with this change?

JuanCarlosM1:
I tested it in ultra wide resolution (21: 9) and it doesn't fix the bug of that resolution however it stays the same as it is without it.

I haven't tested it at all because my PC is not very good.

slashiee:
Maybe you can try this to see if it works with Ultrawide? I forgot what we do for these cases where we need the text to stay the same aspect ratio.

```
[TextureRedefine] #Pictograph
width = 1600
height = 912
formats = 0x019
overwriteWidth = ($height/$gameHeight) * 1600
overwriteHeight = ($height/$gameHeight) * 912

[TextureRedefine] #Pictograph
width = 1600
height = 900
formats = 0x019
overwriteWidth = ($height/$gameHeight) * 1600
overwriteHeight = ($height/$gameHeight) * 900
```

Also, requesting anyone with a good PC to test higher resolutions with this change.

slashiee:
i just pushed the that change to this PR, it should fix UW resolutions as well

PR #442, opened 2020-05-14
Added "Hyrule Warriors Coop Aspect Ratio" gfx pack

MelonSpeedruns:


Resolutions/HyruleWarriors_Coop_AspectRatio/patches.txt
```diff
@@ -0,0 +1,7 @@
+[ProjectZ_219v208]
+moduleMatches = 0xE159AE91
+
+0x1009609C = .float $ratio
+0x10096124 = .float $ratio
+0x10096948 = .float $ratio
+0x10096974 = .float $ratio
```
slashiee, line 7: Are you sure we need to change all these values? I feel like two of them would be used just for the co-op mode.

MelonSpeedruns: I didn't check, but I got those values from the resolution pack, and that uses all 4 of them.

slashiee: Did you resolve it because you checked to make sure all 4 changes were necessary?

slashiee:
@MelonSpeedruns Are you sure that we need to modify all the aspect ratios for this pack?

slashiee:
I've just noticed numerous issues with this pack, such as location and folder naming. I feel like it's best to have this be added as part of the version 5 pack instead, where we have more control with overriding stuff.

PR #444, opened 2020-05-24
[Paper Mario: CS] Fix the invisible characters in Cutout Mode

RedHalted:
This is a small fix to render Mario and the flag over the background during cutouts sequences. 

[Example](https://gfycat.com/soulfuldistortedcaribou).


Crementif:
That's a really good find, thank you so much for submitting your workaround!

Btw, would you want your name to be credited in the graphic pack itself? It's something we do now, but you're free to let it go uncredited. Everything else looks good to merge.

RedHalted:
Oh, sorry I didn't know. I added the credits to the description :+1:

PR #445, opened 2020-05-24
Add 007 Legends

williansti:
add resolution pack


Resolutions/007 Legends/rules.txt
```diff
@@ -0,0 +1,112 @@
...
+version = 4
+
+[Preset]
+name = 1920x1080 (Default)
+$width = 1920
+$height = 1080
+$gameWidth = 1920
+$gameHeight = 1080
+
+// Performance
```
slashiee, line 15: We try to use # (ex. `# Performance`) for comments

PR #448, opened 2020-05-28
Cleanup Child of Light pack

icup321:
Cleaned up some of the text for Child of Light, and also Cemu doesn't like "&" for some reason so I changed it to "and" for Pac-Man.


Resolutions/ChildOfLight_Resolution/rules.txt
```diff
@@ -108,26 +107,28 @@ formats = 0x01a
 overwriteWidth = ($width/$gameWidth) * 512
 overwriteHeight = ($height/$gameHeight) * 512
 
-[TextureRedefine] # Half-res
+[TextureRedefine] # Half-Res
+width = 640
+height = 368
+tileModesExcluded = 0x001
+overwriteWidth = ($width/$gameWidth) * 640
+overwriteHeight = ($height/$gameHeight) * 368
```
Crementif, line 115: Not all games use padded textures, is there a reason why you added this? Because if it's not used by the game I'd rather not define it at all.

Resolutions/ChildOfLight_Resolution/rules.txt
```diff
@@ -108,26 +107,28 @@ formats = 0x01a
...
+
+[TextureRedefine]
 width = 640
 height = 360
-formats = 0x001
```
Crementif, line 114: Defining a format prevents unwanted textures being upscaled and makes it easier to work on in the future since you keep track of what each TextureRedefine does.

Resolutions/ChildOfLight_Resolution/rules.txt
```diff
@@ -108,26 +107,28 @@ formats = 0x01a
...
 overwriteHeight = ($height/$gameHeight) * 240
 
 [TextureRedefine]
-width = 448
+width = 426
 height = 240
-formats = 0x01a
-overwriteWidth = ($width/$gameWidth) * 448
+overwriteWidth = ($width/$gameWidth) * 426
 overwriteHeight = ($height/$gameHeight) * 240
```
Crementif, line 134: Personally think these are mostly just preference things, do I do agree that putting the `# Screen gradient` on the right might be a little cleaner. 🤷

Resolutions/ChildOfLight_Resolution/rules.txt
icup321: Well, I did notice the game loads 640x368 textures on memory based on my texture dumps, but I can't confirm if it actually renders them, but I thought it would be safer to add them.

Resolutions/ChildOfLight_Resolution/rules.txt
icup321: I am aware of this, but I didn't notice any visual bugs when removing this, so I thought it was unnecessary to keep this line. Slashiee doesn't seem to add formats in his packs unless it's really necessary. Perhaps a # could be added in front?

Resolutions/ChildOfLight_Resolution/rules.txt
icup321: Yes, I do prefer having them ordered from biggest to smallest textures because it makes it easier to search for especially when there are a lot of textures being up-scaled. If you don't like all the other changes I made to this pack, I can easily just revert them. I just wanted to clean this pack specifically since I based one of my packs off of this one since Rayman Legends runs on the same engine.

Resolutions/ChildOfLight_Resolution/rules.txt
Crementif: You can also enable the View Texture Cache Warnings in the Debug->Logging menu to see that there's no problem where a padded layer needs to be added.

Resolutions/ChildOfLight_Resolution/rules.txt
Crementif: It's more work to define it while not always being that necessary. Unless it helps the game in some way, I don't see why you would remove it.

Resolutions/ChildOfLight_Resolution/rules.txt
Crementif: I think reverting the `formats = 0x1a` here would resolve this issue.

Crementif: I think it's just a personal preference thing. I like to make my graphic packs very orderly and very defined, but I'm not saying that it's the right way to make graphic packs.

Resolutions/ChildOfLight_Resolution/rules.txt
```diff
@@ -89,7 +89,6 @@ $gameHeight = 720
 [TextureRedefine] # TV Resolution
 width = 1280
 height = 720
-formats = 0x011,0x01a,0x001
```
Crementif, line 92: Oh, I forgot to add this one as a comment. Same thing as the other formats. Sorry!

icup321: Okay I've added back that one as well, that should be it, right?

Crementif: I still don't agree with adding the padded half-res TextureRedefine and you didn't re-add all the format filters 😅

icup321: Okay, that's everything. Can this be merged now?

Crementif:
Thanks for all the work 😅

PR #452, opened 2020-07-03
Additional Splatoon Patches

IVy-The-Wusky:
The "Better Squid Sisters Lighting" patch just removes the normal news lighting so it's set to what it would normally look like in that position in the plaza, making it look "somewhat" better. This also includes a patch the removal of the 2D news layer, and a special swapper, all made by me.

### DIFF
--- a/Mods/Splatoon_BetterSquidSistersLighting/patches.txt
+++ b/Mods/Splatoon_BetterSquidSistersLighting/patches.txt
@@ -1,0 +1,4 @@
+[Gambit272]
+moduleMatches = 0xF7A78809
+
+0x100F4B18 = nop # Disables "MainNews" lighting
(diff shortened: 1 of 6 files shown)


slashiee:
Out of curiosity, what does the special swapper do? Does it swap the model?

IVy-The-Wusky:
Essentially, yes, when I swapped the Killer Wail with the Rainmaker it looked like this:
image

IVy-The-Wusky:
The special swapper patch is pretty unstable atm, it's better to keep it out of the repo for now so I can test everything on it.

slashiee:
Can you at least make it clear that it swaps the model, and make sure it doesn't cause any issues with online play?

IVy-The-Wusky:
It just swaps the model, and I've tested it online and I haven't had any issues, but it seems that it only lets you swap the Rainmaker over something which makes it pretty useless :/

IVy-The-Wusky:
Although it still could be useful, I will add a note and make some little changes to adapt it to this

PR #493, opened 2021-01-17
[BotW] DivineLaserBeams: Make transparency and saturation settable + UX improvements

Fs00:
Let me join the excitement for BotW v6 packs 😛 

Since the first version of the pack made by Kiri, it was possible to change beams saturation and transparency by manually editing the shaders.
When I ported the pack to version 3, it wasn't possible to have preset categories, so I couldn't add a way to change those two options in the UI. But now we can, so I did it! 🎉 
Users can now choose between 100%, 75%, 50% and 25% for color saturation. For transparency is the same, plus a 10% option (for those who prefer barely visible beams).
I've also reworded the enable/disable beams preset to be a little clearer and hidden all color options if beams are disabled.

PS: there is third settable constant in the shaders, which is "value". I didn't make it editable from the UI because it seemed to behave like transparency.


Crementif:
Seems great. Still not sure if Enabled and Disabled don't do the job (which were initially chosen as they're more universal and those terms are used everywhere), but Show and Hide are fine too I suppose.

Fs00:
I changed them because the phrasing "Hide Laser -> Enabled" and "Hide Laser -> Disabled" was a bit difficult to grasp at a first glance (at least for me), especially when you have a double negation.
"Show Laser -> Yes" and "Show Laser -> No" is more direct and easy to understand imho.

Thanks for your work on these packs!

PR #513, opened 2021-04-07
[Sonic Boom] Disable Version String patch

HyperBE32:
Hides the version string that lingers around on the menus in the patched version of the game.


ActualMandM (approved):
image

PR #516, opened 2021-04-09
[Sonic Boom] Disable SSAO

HyperBE32:
A simple patch that eradicates SSAO from the game by skipping the parameter setup.


src/SonicBoomRiseOfLyric/Mods/DisableSSAO/rules.txt
```diff
@@ -0,0 +1,6 @@
+[Definition]
+titleIds = 0005000010175B00,0005000010177800,0005000010191F00
+name = Disable SSAO
+path = "Sonic Boom: Rise of Lyric/Mods/Disable SSAO"
```
ActualMandM, line 4: It might be better to put this under Enhancements, which is where graphical modifcations would typically go for other games

PR #548, opened 2022-02-15
Update rules.txt

victoria4dx:
Added 32:9 aspect ratio resolutions


Crementif:
Sorry that it took so long! I slept on my maintainer duties for a bit. Thanks for adding it!

PR #564, opened 2022-08-02
9999 Stamps Limit Cheat graphic pack

Waikuteru:


Crementif:
Awesome! Thanks for contributing this new cheat :) I slightly adjusted the description (and name) to make it clearer this applies to map stamps and did some other adjustments. Hope the credits are still good.

PR #577, opened 2022-11-22
[BOTW] Add 3:2 resolutions

### LINKED ISSUE #576: [BOTW] 3:2 Aspect Ratio Support
I have a Framework laptop with a 3:2 2256:1504 display. Currently there is no option in the BOTW graphics selection to allow for a 3:2 aspect ratio. I think adding one would just be a simple case of expanding the options in https://github.com/ActualMandM/cemu_graphic_packs/blob/master/src/BreathOfTheWild/Graphics/rules.txt, and it would make anyone using a Framework laptop's experience a lot better.

**Additional context**
Screenshot from 2022-11-21 18-23-50
See lack of 3:2 aspect ratio option.

j5155:
I added a 3:2 aspect ratio, and resolutions for it.
Closes #576

### DIFF
--- a/src/BreathOfTheWild/Graphics/rules.txt
+++ b/src/BreathOfTheWild/Graphics/rules.txt
@@ -90,6 +90,13 @@ $aspectRatioHeight = 4
 $showUltrawideOptions:int = 0
 $ultrawideHUDMode:int = 0
 
+[Preset]
+name = 3:2
+category = Aspect Ratio
+$aspectRatioWidth = 3
+$aspectRatioHeight = 2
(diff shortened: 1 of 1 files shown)


intra2:
did you forget to upload the changed rules.txt?

j5155:
...yes. Sorry, one moment!

j5155:
Okay, that should be good. There are a few minor graphical issues with things overlapping because it's less wide then the game expects, but it's definitely playable and a lot nicer of an experience, and the expertise needed to fix that is more then I have.

PR #583, opened 2022-12-31
[Xenoblade X] New blade metal count mod + some other stuff

intra2:
New blade metal mod to set you current blade metal count.

Global nemesis offline has a new cheat option to disable the BLADE Medal mechanic

some typo fixes

new cheat option in collectables catch range

reorganized the HUD and Graphics mods because they were organized badly


intra2:
the new blade metal mod doesn't work as intended. converting to draft till I can figure it out

intra2:
Fixed the blade metal mod. works properly and as intended now. also found a way to remove the 9999 blade metal cap (even though its almost impossible to hit)

intra2:
unless I figure out how to start offline nemesis with 0 medals everything is ready!!

intra2:
> unless I figure out how to start offline nemesis with 0 medals everything is ready!!

I just figured out how to do this so now everything should be ready

intra2:
this *should* be everything

PR #586, opened 2023-01-17
[Angry Birds Trilogy] Add 60FPS

LolHacksRule:


intra2:
patches.txt should be renamed to patch_angrybirdsfps.asm then everything will be good on your end.
would just need to wait for crem or actual to take a look at the pr's

PR #599, opened 2023-06-25
Some additions to the Day Length mod for BotW.

chrispurnell:


src/BreathOfTheWild/Mods/DayLength/rules.txt
```diff
@@ -29,99 +31,169 @@ name = 15 seconds
...
+name = Scaled with day length
+category = Cloud Speed
+condition = ($timeCycleMode == 1)
+default = 1
+$cloudMultiplier = 0
+
+
+[Preset]
+name = 5 minutes
+category = Clock Interval
```
Crementif, line 131: Just to be nitpicky, but I think this name could be a bit more descriptive perhaps. Ultimately up to you, but something like Clock Update Frequency, Clock Increment Rate or something else could work.

src/BreathOfTheWild/Mods/DayLength/rules.txt
```diff
@@ -29,99 +31,169 @@ name = 15 seconds
...
+condition = ($timeCycleMode == 1)
+default = 1
+$clockAdjust:int = 1
+
+[Preset]
+name = 1 minute
+category = Clock Interval
```
Crementif, line 138: Same as above

src/BreathOfTheWild/Mods/DayLength/rules.txt
```diff
@@ -29,99 +31,169 @@ name = 15 seconds
...
+[Preset]
+name = Scaled with day length
+category = Cloud Speed
+condition = ($timeCycleMode == 1)
+default = 1
+$cloudMultiplier = 0
+
+
+[Preset]
+name = 5 minutes
```
Crementif, line 130:
```suggestion
name = Increment Clock By 5 Minutes (Default)
```
I think this'd be a bit clearer for a user. The dropdown names can be pretty long.

src/BreathOfTheWild/Mods/DayLength/rules.txt
```diff
@@ -29,99 +31,169 @@ name = 15 seconds
...
+category = Clock Interval
+condition = ($timeCycleMode == 1)
+default = 1
+$clockAdjust:int = 1
+
+[Preset]
+name = 1 minute
```
Crementif, line 137:
```suggestion
name = Increment Clock By 1 Minute
```
Could go with every minute too, but I think it's easier to parse if you use the same formatting for both. Just my thoughts.

Crementif (commented):
These are some cool improvements, nice work! I tried it in-game and it seems to work great.

I do have some minor nitpicks to make things a bit more descriptive, maybe. Let me know if you agree with them or feel free to do your own thing. After that I'll merge it.

src/BreathOfTheWild/Mods/DayLength/rules.txt
chrispurnell: I was unsure about what to put for a lot of that label text.  Your suggestions all seem good.
The other things I was unsure about was the exact values for the blood moon presets or if I should add other cloud speed presets.

Crementif: I think with the amount of presets and types of presets I'd generally just go with what makes sense to you and you see someone using.

Crementif:
Looks good to me, you want me to merge it?

chrispurnell:
Yes please.

PR #602, opened 2023-07-05
[Zelda Wind Waker] Why not support JP game workaround？

KiesProqwq:
One day i play WindWaker,suddenly i find 3 workaround titleIds is not suport the JP version,so I add a tilelds in that.

titleIds = 0005000010143400,0005000010143600
titleIds = 0005000010143400,0005000010143600,0005000010143500


Crementif:
The reason why it didn't support it is because each region has their own executable, so any patches require to be manually adjusted. However, I've spend some time tonight to make it work for the japanese version too.

PR #610, opened 2024-03-10
Fix Mario kart 8 lines bug

### LINKED ISSUE #573: [Mario Kart 8] When enabling the graphics mod weird line glitches

goeiecool9999:
Fix #573
Before:
Screenshot_2024-03-10_23-15-32
After:
Screenshot_2024-03-10_23-13-32
Here you can see what the textures look like in RenderDoc
Screenshot_2024-03-10_23-13-12
Screenshot_2024-03-10_23-15-51

Oddly enough this only happens in OpenGL. Unless there's some rationale as to why this specific format needs to be changed that I'm unaware of it's probably best to remove this format change

### DIFF
--- a/src/MarioKart8/Graphics/rules.txt
+++ b/src/MarioKart8/Graphics/rules.txt
@@ -1012,12 +1012,6 @@ formats = 0x806
 overwriteFormat = 0x820
 
 [TextureRedefine]
-width = 1280
-height = 720
-formats = 0x816
-overwriteFormat = 0x820
-
-[TextureRedefine]
 width = 800
 height = 450
 formats = 0x01a


Exzap:
> Unless there's some rationale as to why this specific format needs to be changed that I'm unaware of it's probably best to remove this format change

It's there to avoid color banding, but I don't know under which circumstances it would appear. Seems like this texture is used only for bloom or glow effects?

Anyway, I dug into this deeper on the Cemu side and it turns out there were bugs with the format overwrite feature on both OpenGL and Vulkan backends. On the latter it would mostly just be ignored and thats why there are no glitches. Commit with fixes will be pushed soon. I don't think we need to revert this PR unless we figure out what the removed rule does and if its even beneficial.

PR #612, opened 2024-03-16
[Wind Waker HD] Added 16:10 resolution

Ryce-Fast:
This patch adds 16:10 resolutions to the game Zelda Wind Waker HD. This will allow you to play on Steam Deck. This patch fixes https://github.com/cemu-project/cemu_graphic_packs/issues/607.


Crementif:
Is it worth reopening this? I'm kinda hesitant on adding support for weird aspect ratios, especially for Wind Waker HD since it has a weird resolution upscaling setup due to the nature of the game's rendering. But if you've tested it and don't see anything off then we can merge this.

PR #618, opened 2024-03-28
MK8: Fix cup screen black lines

goeiecool9999:
Fixed an issue where on the results screen of a Grand Prix race there would be a black bar on the scoreboard.
before:
screenshotbroken
after:
Screenshot_2024-03-28_16-40-33
The normal rendering of the scoreboard uses a texture of 576x64 and copies it to a smaller texture of 560x64:
image
image
However the resolution graphic pack puts the source image at the same resolution as the destination, so the image copy doesn't cut off the black/undefined part which shows in the final output.


Crementif:
Yeah, that definitely seems like a copy-paste mistake haha!

PR #626, opened 2024-05-18
[Paper Mario Color Splash] Adding a graphic pack to be able to modify gamepad resolution, to make it easier to play using the Off-TV play option

bobleblob123:


Crementif:
So does this allow you to see the gamepad screen or is this more of a suggestion? Can you post a screenshot.

bobleblob123:
> So does this allow you to see the gamepad screen or is this more of a suggestion? Can you post a screenshot?

I can later today but the way it works is allowing you to specify a resolution for the gamepad screen as previously the resolution options did not alter (the textures at least) gamepad resolutions for Color Splash

bobleblob123:
> So does this allow you to see the gamepad screen or is this more of a suggestion? Can you post a screenshot?

would you prefer a screenshot of the graphics pack itself in cemu or what specifically should I send?

bobleblob123:
actually this does not work currently, so it will be closed for now

bobleblob123:
I have concluded that it is placebo and technically doesn't do anything

PR #628, opened 2024-06-02
Wind Waker HD Graphic Pack rewrite

Ryce-Fast:
I partially rewrote all the graphics packs related to WWHD.

- Moving all graphics packs to src
- Aspect Ratio, resolutions, light resolution and shadows in Graphics
- Texture LOD and contrasty in Enhancements
- Added option for lights resolution
- Adding 32:9 and 48:9 resolution (No issues with the UI)
- Adding 10K resolution for 16:9
- PictoBoxFix in Cheats
- If you have ideas for changes, let me know!

Graphics:
image

Enhancements:
image

21:9
Capture d'écran 2024-06-03 181904

32:9
image

48:9
image


Crementif:
This is an interesting project, so thanks for contributing it! We tried to migrate all graphic packs to V5 a few years ago over at the now privated https://github.com/ActualMandM/gfx_pack/ (@ActualMandM do I have permission to fork and publicize it?), but we ended up halting that effort when we realized that the testing and potentially breaking the configuration was a lot of effort.

If you want to continue that effort, I think its worth it to review the graphic packs we made there. I think most of them are done, just untested. I'll try and see if I can be more proactive with reviewing these PRs, but I think the current graphic pack layout is acceptable for most people too so I do question the effectiveness.

Here's the graphic pack I made for Wind Waker HD back then.
[WindWakerHD.zip](https://github.com/user-attachments/files/15779516/WindWakerHD.zip). Maybe you can use that as a base? I think the naming and categories are just a lot better then the graphic packs you used as a base. In my graphic pack I did separate the TV and Gamepad Resolution, but in hindsight that caused a lot of bugs. So if you use it, merge those options.

Some other notes:

- Can you test whether the pictobox quests works without the workaround graphic pack. I've been told it has been fixed in Cemu. The pictobox fix should then also be named just "Allow Any Pictobox Image" or something like that.
- I don't like using version=7. The graphic packs don't use any features from anything above 7, so using version=5 is better since that's compatible with many more versions.
- You should use `default = 1` for the default options. Also, you should be suffixed with a `(Default)` text. This way its easy for users to switch back to default settings since you're now merging graphic packs and some people might not want to use it.
- I think the Contrasty option should be kept separate since it overwrites the game's anti-aliasing shaders, which we don't want to use as a default.

ActualMandM:
I can make it public again, I only privated it since I didn't want people to end up getting outdated packs back then.

PR #638, opened 2024-07-26
[Skylanders Spyro's Adventure] Move new pack to /src

SuperSamus:
#629 put the new version of the pack in the root directory of the project, instead of modifying the pack already present in `/src`.
I don't know if that was a mistake or if it was intentional, so... either merge or close, depending on which.


Crementif:
Oh whoops, that definitely was a mistake. Thanks for the fix!

PR #639, opened 2024-07-26
Skylanders: Fix infinite fall softlock at high FPS

SuperSamus:
Fixes one of the issues caused by higher framerates in the Skylanders saga: #615.

The reason why the character was respawning during long falls is because of a failsafe meant in case your character is stuck falling between geometry.
The game determines whether you are stuck with a simple method: it calculates the squared distance of your character from the position of the previous frame, then checks whether it's lower than 20². If that's true for 3 seconds, then it teleports you.
But if the framerate is higher, your distance is going to be lower than normal, and the game will consider you stuck even when falling at top speed.
Ironic...

The fix simply consists in overriding the load of 20.0, and load `20×30FPS/targetFPS` instead.
(Ideally you would multiply by `deltaTime` instead of dividing by `targetFPS`, but since it's not already loaded in a register in that part, not bothering.)

Before:

https://github.com/user-attachments/assets/c6215296-79bd-4a04-810e-b9f35508fbd0

After:

https://github.com/user-attachments/assets/74097add-1df3-4f8a-92f8-3463367f13d8

Credits:
- Me for understanding and patching the issue on the Wii version of Swap Force.
- @WinnerNombre for finding the addresses for the Wii U games. (And for the videos.)

This is currently a draft, as this is only one version of one game, so not worth merging now. More games and versions will be added later.
I'm leaving this for documentation and for reference.

Other things for reference:
- For games that have debug symbols, the function is named `updateStuckFallingCheck`.

### DIFF
--- a/src/SkylandersSuperChargers/Mods/FPS/patch_FixInfiniteFall.asm
+++ b/src/SkylandersSuperChargers/Mods/FPS/patch_FixInfiniteFall.asm
@@ -1,0 +1,51 @@
+[SSC_FixInfiniteFall]
+moduleMatches = 0xccf357b3, 0x1407432e, 0x3a1208f7, 0x21fc6048, 0x3e598867, 0xcf77bcd5, 0x25a11655
+.origin = codecave
+
+const_stuckFallingDistanceThreshold:
+.float 20*30/$targetFPS
+
+_getStuckFallingDistanceThreshold_f13:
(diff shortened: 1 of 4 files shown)


SuperSamus:
Both versions of Swap Force are here!

Crementif:
That's cool! Is this still intended to be a draft pull request, or is it mergeable. I'll trust you to say whatever is better for the users.

PR #645, opened 2024-10-02
[N64] Better resolutions

Ryce-Fast:
I put the resolution up to 8K instead of 4K.
Tested on Ocarina of Time and Majora's Mask


intra0:
was there any problem with how the resolution pack was handled previously??

PR #651, opened 2024-11-19
[Rise of Lyric] Note Knuckles jump glitch in Fix Pause State Reset

ActualMandM:
Adding as a note after watching a friend play the game and question why the glitch wasn't working, then seeing this pack enabled lol


Crementif:
Oh, sounds good. Hope you're doing well, and thanks for the PR!

PR #652, opened 2024-11-21
[BOTW] Make Option-Orders consistent

### LINKED ISSUE #591: [Zelda Breath of the Wild] Inconsistent Settings-Menus in Graphicpacks-Section
This may seem like a stupid thing, and it isn't something very bad, but i could find this "issue" in a lot of games.

Also, i don't know if this affects other Games, i could only reproduce it on botw, because it's the only game i've got dumped.

While some settings, like DOF and Anisotropic Filtering have their "high/enabled" setting at the top:
image

On the other hand, there are some that have this behavior completely reversed:
image

This leads to some very confusing situations, for example it is completely random in the Enhancements-Category:
image

While this isn't something urgent, it'd still be nice if it would be addressed at some point.

lunatic-gh:
For a consistent UX-Experience, Presets considered as "LOW" or "OFF" should always be put on top, while "HIGH" or "ON" should be on the bottom, eg `low -> medium -> high -> max`

The BOTW GraphicPacks category seems like it could not decide, and just screwed the order entirely, leading to some doing it like this, and some like that.

This PR re-orders the options from the Enhancements and Graphics categories. I left others untouched (eg Cheats and Mods), because they apparently follow the `default -> non-default` order. 

Would close #591

### DIFF
--- a/src/BreathOfTheWild/Enhancements/rules.txt
+++ b/src/BreathOfTheWild/Enhancements/rules.txt
@@ -126,14 +126,14 @@ $preset:int = 0
 # Depth of Field
 
 [Preset]
-name = Enabled
-category = Depth Of Field
-default = 1
-
-[Preset]
 name = Disabled (no performance gain)
 category = Depth Of Field
 $disableDOF:int = 1
+
+[Preset]
+name = Enabled
+category = Depth Of Field
+default = 1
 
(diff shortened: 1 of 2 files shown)


Crementif:
Yeah, this seems like it'd make a lot of sense. Thanks for making a pull request!

PR #655, opened 2024-11-29
[MK8/Splatoon/Minecraft] Add overrides for the peer-to-peer/multiplayer ports

ashquarky:
Wii U games have pretty dated netcode that can struggle with modern networks, leading to 118- errors when playing online. Nintendo's recommendation is to place the game in the DMZ, effectively forwarding every port to it. This isn't really viable for most users.

These patches allow the user to specify a port number that they would like the game to use for all peer-to-peer communications. Then, if the user forwards that UDP port in their router, it's equivalent to the DMZ/"NAT Type A" case and they should get better connectivity. Pretendo plans to ship similar patches to console players so we can maximise the number of port-forwarded users in the online population (the port number will be automatically chosen on hardware - I couldn't work out a good way to do this in Cemu though).

Minecraft is on a much newer NEX library version which is why it has a substantially different patch.

This is my first time contributing Cemu graphic packs, feel free to let me know if I misunderstood which folder to put these in or how to write the description or the like and I'll happily fix it.

### DIFF
--- a/Workarounds/MinecraftWiiUEdition_NetworkPort/patch_port.asm
+++ b/Workarounds/MinecraftWiiUEdition_NetworkPort/patch_port.asm
@@ -1,0 +1,14 @@
+[MinecraftWiiU_ForceMultiplayerPortNumber_v688]
+moduleMatches = 0x867317DE
+
+# GetRandomPortRangeMin
+.origin = 0x03579530
+li  r3, 0
+ori r3, r3, $port
+blr
+
+# GetRandomPortRangeMax
+.origin = 0x0357953c
+li  r3, 0
+ori r3, r3, $port
+blr
--- a/Workarounds/MinecraftWiiUEdition_NetworkPort/rules.txt
+++ b/Workarounds/MinecraftWiiUEdition_NetworkPort/rules.txt
@@ -1,0 +1,59 @@
+[Definition]
+titleIds = 00050000101D9D00,00050000101D7500,00050000101DBE00
+name = Multiplayer network port
+path = "Minecraft: Wii U Edition/Workarounds/Multiplayer network port"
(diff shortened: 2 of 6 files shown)


Workarounds/MarioKart8_PortNumber/patch_port.asm
```diff
@@ -0,0 +1,7 @@
+[MarioKart8_v81]
```
Crementif, line 1:
```suggestion
[MarioKart8_ForceMultiplayerPortNumber_v81]
```
This would make it easier to identify in the log.txt (when graphic pack logging is enabled under Debug->Logging) and PPC debugger.

Workarounds/Splatoon_PortNumber/rules.txt
```diff
@@ -0,0 +1,59 @@
+[Definition]
+titleIds = 0005000010176900,0005000010176A00,0005000010162B00
+name = Multiplayer port
+path = "Splatoon/Workarounds/Multiplayer port"
+description = Forces the multiplayer port to be 58008.
```
Crementif, line 5: I'm sure no one will be caught off guard by this, but probably best if this matches the other packs.

Crementif (changes requested):
Seems like some nicely written and useful patches! There's only some minor nitpicks that would be nice to have changed:

Ideally we would put the Mario Kart 8 and Splatoon graphic packs into the src/MarioKart8/Workarounds folder respectively. The Minecraft graphic packs haven't been updated to use the newer layout/features yet, so they're correct already.

I think it might be a bit unclear for less technical minded people on what these graphic packs do and whether they should enable or disable them. I think using wording such as network port or multiplayer network port would make it easier to google. And calling it port forwarding instead of just forwarding a port. Also, adding a line at the end about why they might need this graphic pack like e.g. "This may resolve issues while connecting to online services." would be a helpful addition for these people.

ashquarky:
Thanks for the review! Should be looking better now.

Crementif (approved):
LGTM

Issue #575, opened 2022-11-12
[WWHD/TPHD] Request adding anisotropic filtering


NineKain:
For Wind Waker it's not that bad, but for Twilight Princess it would be a godsend

1112e:
just curious, can this be done in driver settings/ nvcp?

intra0:
pr #647 will fix this by adding the graphics packs. this issue can safely be closed after that pr gets merged

PR #657, opened 2024-12-12
[XCX] Fixes up sugested changes from last PR, updates max money mod | add *a lot* of credits comments

intra0:
Implements the suggested changes from the last PR (typo fix in brightness old, change `float` to `double` in FXAA, remove some `:`)
Renames moon jump to super high jump because the name moon jump is kinda misleading
the max money in the max money mod is now editable.
changed the rules.txt for ticket multiplier a bit.
new readme.txt document in the Character level mod.
Character level can now also edit character exp.
you can now edit individual stats in the enemy stats mod.

Then adds the credits comment to every file (thats why its a 103 file count)
also deleted some unnecessary `int $mod` declarations because I didn't quite get gfx packs when I added those.
ik the 103 file count looks scary but its a lot less than it looks

(only jump height, max money, character level, Enemy stats, & ticket multiplier have *real* changes)


intra0:
you also forgot to close #575 from the last PR

src/XenobladeChroniclesX/Mods/PhysicsSuperHighJump/rules.txt
```diff
@@ -1,8 +1,8 @@
 [Definition]
 titleIds = 0005000010116100,00050000101C4C00,00050000101C4D00
-name = "Jump higher"
```
intra0, line 3: the `(Vanilla values are 1.0 for both)` isn't needed in the description because the x1 option for both distance and height have `(default)` in the name

src/XenobladeChroniclesX/Enhancements/AntiAliasing/rules.txt
```diff
@@ -8,9 +8,9 @@ version = 6
 [Default]
 $Preset:int = 0
 
-$subPix:float = 0.75
-$edgeThreshold:float = 0.166
-$edgeThresholdMin:float = 0.0833
+$subPix:double = 0.75
+$edgeThreshold:double = 0.166
+$edgeThresholdMin:double = 0.0833
```
intra0, line 13: I have tested :double, the mod works just the same, there is a noticable difference between native (mod off), fxaa, and aa disable.
so working exactly as intended

src/XenobladeChroniclesX/Mods/PhysicsSuperHighJump/rules.txt
```diff
@@ -1,8 +1,8 @@
...
-path = "Xenoblade Chronicles X/Mods/Physics/Jump to the Moon!"
-description = Increase horizontal and vertical velocity of jump (Vanilla values are 1.0 for both).
+name = "Super High Jump!"
+path = "Xenoblade Chronicles X/Mods/Physics/Super High Jump!"
```
intra0, line 4: the (Vanilla values are 1.0 for both) isn't needed in the description because the x1 option for both distance and height have (default) in the name

src/XenobladeChroniclesX/Mods/PhysicsSuperHighJump/rules.txt
```diff
@@ -1,8 +1,8 @@
...
+description = Increase horizontal and vertical velocity of jump.
```
intra0, line 5: the (Vanilla values are 1.0 for both) isn't needed in the description because the x1 option for both distance and height have (default) in the name

intra0:
everything here looks good to me

intra0:
Looks good to me rn

intra0:
sperating the credits comment into a seperate commit

PR #658, opened 2024-12-12
[Shantae Half Genie Hero] New Graphics Pack 

intra0:
updates the old v2 pack to v7 so it works on modern versions of cemu

the old v2 pack can be found here:
https://github.com/cemu-project/cemu_graphic_packs/blob/939a478799c13c5ed03e8decbf952f2a7ad02366/Source/ShantaeHGH/rules.txt


intra0:
I will also be adding a `#Credits:`  comment to every mod (again throughout multiple PR's)
for some mods it may not be necessary because the credits are already in the description but I'm still gonna add a credits field anyway just to keep every mod stylistically consistent

I am getting the credits by looking into the commit history for the specific game mods

intra0:
Almost Forgot screen shots. oops

2160p:
Screenshot_2024-12-12_14-17-42

1080p:
Screenshot_2024-12-12_14-19-30

360p:
Screenshot_2024-12-12_14-20-42

PR #661, opened 2024-12-14
[MK8] clean up and New Mods

intra0:
adds v1 support to 60fps and resolution pack
adds the credits comment to every pack
adds bloom pack from M&&M's v5 repository

Adds 7 new save file cheat mods


intra0:
Oh, @ActualMandM in your repository you have a FXAA mod for MK8?
https://github.com/ActualMandM/gfx_pack/blob/master/src/MarioKart8/Enhancements/FXAA/rules.txt

this is the only mod in your repository I dont know what to do with
the Anti-aliasing stuff in the graphics mod is called FXAA, and the graphics mod removes AA, but I read that MK8 doesn't have AA, and basically I am just very confused about the state of anti-aliasing in Mario Kart 8

Do you think your MK8 FXAA pack should be updated to v6/7 and added, and if so would there need to be any changes to the current resolution mod or?? basically whats the deal with ~~airplane food~~ Mario Kart Anti-Aliasing?

.
.
Every other mod in your repository I have already updated to v6/7 and will slowly be merging with this main repository
so thanks for the help

ActualMandM:
MK8 does use FXAA, but only on character select. IIRC the main graphic pack patches the code to also apply when in a race, so the pack from that repo is redundant

ActualMandM:
Oops, turns out the code in the main pack is for disabling FXAA in character select instead, sorry for misremembering

Anyways the pack is a post processing effect. It's been a while since I've really looked much into packs so this is more a question for @Crementif or the original pack creator @getdls

src/MarioKart8/Graphics/patch_DisableFXAA.asm
```diff
@@ -1,10 +1,11 @@
-[MK8_DisableFXAA_V42]
```
intra0, line 1: The patch is called disable FXAA though??
I'm still confused

src/MarioKart8/Cheats/TrophyCheat/rules.txt
```diff
@@ -0,0 +1,171 @@
+[Definition]
+titleIds = 000500001010EC00,000500001010ED00,000500001010EB00
+name = Gold Trophy / 3 stars
+path = "Mario Kart 8/Cheats/Grand Prix Trophies"
+description = Cheat in Grand Prix Trohpies!|Must complete or quit a Grand Prix to save changes.
```
intra0, line 5: If you have a better sugestion for the `Must complete or quit a Grand Prix to save changes.` line
or Better names for `Change EVERYTHING (200/mirror/150/100/50)` `Change Entire Engine Class` `Change Specific Cup (activates by cursor moving onto 50cc)`
please let me know. I think they're fine but feels like they could be worded better

getdls:
I could have sworn the patch AA to in-game was working  and thats why  we nuked FXAA?? 
The FXAA pack was mostly a copy-paste from XCX with really low settings to avoid the smearing, while still fixing some of the worst specular edge aliasing.

intra0:
@getdls tested it and could see no difference between having it enabled vs disabled.
Was using OpenGL

getdls:
Thnx, for checking.. and good work on the other stuff 
Mandela effect I guess :/  Trust,  but verify! ;)

intra0:
gonna add the new mods and .asm changes as their own commit
and add the credits comment on an alt account as its own commit

PR #662, opened 2024-12-14
[Virtual Console] Add missing title ID's, new packs for different systems, & move into src

intra0:
Updates the packs to v6 and moves them into SRC
Adds missing Title ID's so all games will be supported
also adds an 8k option to n64 from #645 but takes no other changes from that PR
I don't think there's a point to any of the changes in #645 and it can be closed without merging
new resolution pack for MSX, and Turbo Graphics 16


intra0:
MSX castlevania at working 8x resolution:

Screenshot_2024-12-17_16-00-11
Screenshot_2024-12-17_16-01-23

intra0:
Turbo Graphics 16 Bomberman '93 working at 6x resolution:

Screenshot_2024-12-17_16-09-00
Screenshot_2024-12-17_16-09-42

PR #669, opened 2024-12-15
[Captain Toad] new AA enable/disable option

intra0:
moves captain toad into src
updates to v6
merges shadows and resolution mod to be consistent with other games
adds an AA enable/disable option


intra0:
Screen shot of the pack working:

1440p, 300% shadows, AA enabled:
Screenshot_2024-12-15_14-53-11

1440p, 300% shadows, AA disabled:
Screenshot_2024-12-15_14-53-52

Exzap:
Why are you increasing the GLSL version to 450? Cemu uses 430 and you shouldn't artificially raise requirements unless there is a good reason to do it.

PR #674, opened 2024-12-31
Skylanders trap team hud and blur filter removel 

Flextape05:
Screenshot 2024-12-31 162626
Screenshot 2024-12-31 163059


Flextape05:
i added motion blur removel

Flextape05:
i added some fixes to the resolution patch the game runs now at the resolution you select before the game would only run at 1440p when selecting 1800p for example

numabello:
sorry guys I don't want to bother you but I don't know what to do and I'm very confused I'm not familiar with programs and I don't understand where I have to put all these codes and I couldn't find the src folder where to put the files you modified, please could you give me a hand ?

PR #681, opened 2025-02-10
[BotW] Use Extreme Quality for FXAA

Ryce-Fast:
FXAA quality at Extreme. Performance remains the same and the result is less pixelated and sharper.

### DIFF
--- a/src/BreathOfTheWild/Graphics/f14bb57cd5c9cb77_00000000000003c9_ps.txt
+++ b/src/BreathOfTheWild/Graphics/f14bb57cd5c9cb77_00000000000003c9_ps.txt
@@ -1,4 +1,4 @@
-#version 420
+#version 430
 #extension GL_ARB_texture_gather : enable
 #extension GL_ARB_separate_shader_objects : enable
 #ifdef VULKAN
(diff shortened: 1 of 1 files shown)


Exzap:
Can you post comparison screenshots? Also did you make sure you were GPU bottlenecked while testing the performance (e.g. by setting the resolution very high) because the shader file has a comment that says 39 is "very expensive"

skalfate:
I was curious as well. Thanks for posting the screenshots. The difference is very visible. I like it. 

Although what if, we did this change as a preset option? As there are laptop users for Cemu?

intra0:
If performance is a concern we could make the FXAA settings user configurable like how the XCX FXAA pack does it.
So a user could chose between the cheaper 14 preset and the more expensive 39 preset.
(or just allow them to pick **all** the presets).
but I do like 39 better than 14 (but I also have a pretty decent computer)

intra0:
best solution would be adding a 4th option "enhanced FXAA (more expensive)" or what ever and allowing the user to pick it
image

Exzap:
> In 1080p the difference was not very visible, but in 1440p it is more. It is written **very expensive** in the shader file except that we must not forget that this technology was unveiled by Nvidia in 2011 and that this shader was supposed to run on the PlayStation 3 and the Xbox 360. With today's components, it is not really a difference anymore. I only noticed a drop of ~1FPS.

The cost is relative to resolution and people play in higher resolution than they did in 2011. But regardless of what your intuition might tell you, it is good to actually check and put in a little extra effort and care that goes beyond "it works on my machine" if you are making something for the community :)

I did benchmark this on a couple of (mid to high end) machines and it decreased FPS only very little. On my main test setup it decreased FPS from 90 to 89, aligning with what others said). What I did notice is that not using FXAA at all gave me 96 FPS. So it seems to be inherently costly regardless of preset. Either way, I think the difference is small enough that we can just merge this as-is without having to make it a separate option. Thanks for the contribution!

PR #682, opened 2025-02-11
[LEGO City Undercover] Resolution Configuration + Texture Sharpness

Ryce-Fast:
**LEGO City Undercover** is a great game but it doesn't have any graphics pack.
So I decided to create one.
Here is a comparison screenshot (The difference is obvious)

Comparison


intra0:
Nice work!

PR #685, opened 2025-02-24
[Skylanders] Fix speed-up above 120 FPS (SSA, SG and STT)

SuperSamus:
When going above 120 FPS on Skylanders: Spyro's Adventure, Giants, or Trap Team, the game would speed up.
This is because Toys for Bob went out of their way to explicitly code this behavior.

Why would they do this, especially on games that are console exclusive?[^1] Truly one of humankind biggest mysteries...

Anyway, this PR patches out this behavior, so that the game will always run at the correct speed (except for the things where the developers forgot to multiply by delta time: those aren't fixed.)

(Also took the opportunity to make small changes to the descriptions of the other games.)

Credits to @WinnerNombre for porting the patch to the Wii U games.

[^1]: Before you say "but SSA is also on PC": that version (along with the PS3 and Xbox 360 versions) runs on a completely different engine, so it doesn't count. Also, it's capped at 60 FPS anyway.

### DIFF
--- a/src/SkylandersGiants/Mods/FPS/patch_dt.asm
+++ b/src/SkylandersGiants/Mods/FPS/patch_dt.asm
@@ -1,0 +1,4 @@
+[SG_dt]
+moduleMatches = 0x62CC63A5
+
+0x02110EF4 = nop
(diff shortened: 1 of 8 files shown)


Crementif:
Looks good to me(rge). Appreciate the humourous description too, thanks for improving the game so much! :)

PR #686, opened 2025-02-25
[xcx] custom fix

intra0:


intra0:
@Crementif this fixes the $custom bug

PR #688, opened 2025-03-04
[MK8] New mods and give old mods v1 support

intra0:
new MK8 mods & adding v1 support to all the .asm files


intra0:
I went and retested every new mod, and also the other mods who's .asm I updated with the v1 support just to make sure there were no glitches, and there were none

Crementif (changes requested):
Oh wow, these are a lot of great new mods! Just some feedback:

* I'm unsure about having a mod that allows one to change your online ranking, even if its easy to edit your save file. Or maybe I'm missing something.

* I also noticed that you recreated the bloom mod, but we actually already have that option integrated into the main Graphics pack, see this file here: https://github.com/cemu-project/cemu_graphic_packs/blob/09207ededf7b7529591e4a0a6b2d5c890d5df5df/src/MarioKart8/Graphics/e90feef2bca6cb2e_00000000000003c9_ps.txt.

* I personally think the `:` you use in the graphic pack options, or the `!` inside the graphic pack names and descriptions make these packs stand out from the other options.

intra0:
> I'm unsure about having a mod that allows one to change your online ranking, even if its easy to edit your save file. Or maybe I'm missing something.

I agree. The race vr has a legitament usecase as 150cc only happens if everyone in the lobby is above 5000 vr
but there is the obvious just cheating usecase.
there's no usecase for battle points, its just there.

the legitament usecase of 5000 vr can actually be accommodate for by obfuscating the code to like 15-20 inctructions, so that it can only set your VR to 5000 (would implement a check to prevent it from lowering VR just incase someone accidentally puts it one)

someone with .asm knowlage could bypass it, but also they could just search for the value in memory and change it to what ever most easily.

I am going to message the pretendo dev's how they feel about this since they are the ones who are gonna have to deal with this. 
they may be fine with it or they may not, and if not then I'll remove the battle mod entierly and change the vr mod to accomodate the 1 real non-cheating usecase, of getting to 5000vr for 150cc (just have the code calculate 5000 over multiple instructions, if somebody can edit the .asm to give them what the want, then they could have direct memory edited it anyway)

so I agree, hold off merging till we hear from the pretendo devs

intra0:
@Crementif its ready to be merged, I messaged the pretendo devs and they're fine with the obfuscated 5000vr mod

image

PR #689, opened 2025-03-04
[XCX] clean up and add new features to mods

intra0:
Adds new features to the Enemy stats, max money, Global Nemesis offline, a x0 option for the exp mult (very useful in combination with the level change mod), a x0 & x1 for the other points mult,

implements suggested style changes from my last XCX PR

Removes unnecessary .int decorations because they bother me

Changes the name of "camera distance mod" to "default camera distance"
people will have to re-apply the mod and thats bad, but its old name is really missleading and it just bothered me a lot.

there may be some unnessisary changed to the .asm styling but I don't feel like reverting every file, and those style changes have at least 1 actual real change (not counting the pure deleting of unnecissary ints)

### DIFF
--- a/src/XenobladeChroniclesX/Enhancements/FancyFX/rules.txt
+++ b/src/XenobladeChroniclesX/Enhancements/FancyFX/rules.txt
@@ -3,9 +3,10 @@ titleIds = 0005000010116100,00050000101C4C00,00050000101C4D00
 name = Fancy GFX
 path = "Xenoblade Chronicles X/Enhancements/Fancy FX"
 description = Pretty blur, de-band sky, re-align cinematic etc, maintained on Nvidia.|Made by getdls.
-version = 6
+version = 4
 
-[Default]
+[Preset]
+name = 1080 -> 1440
 $width = 2560
 $height = 1440
 $gameWidth = 1280
@@ -14,9 +15,6 @@ $internalRes = 0.75
 $dither = 0.15
 $scaleShader = (1440.0/1080.0)
 $scaleBlur = 0.5
-
-[Preset]
-name = 1080 -> 1440
 
 [Preset]
 name = 1440 -> 2880
--- a/src/XenobladeChroniclesX/Enhancements/FancyFX/rules.txt
+++ b/src/XenobladeChroniclesX/Enhancements/FancyFX/rules.txt
@@ -2,7 +2,7 @@
 titleIds = 0005000010116100,00050000101C4C00,00050000101C4D00
 name = Fancy GFX
 path = "Xenoblade Chronicles X/Enhancements/Fancy FX"
-description = Pretty blur, de-band sky, re-align cinematic etc, maintained on Nvidia.|Made by getdls.
+description = Pretty blur, de-band sky, re-align cinematic etc.|Made by getdls.
 version = 4
 
 [Preset]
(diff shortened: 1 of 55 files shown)


src/XenobladeChroniclesX/Mods/BladeMoreTickets/rules.txt
```diff
@@ -37,8 +41,8 @@ name = "Gain more Tickets (x1000)"
 $mult = 1000
 
 [Preset]
-name = "Respawn Global Nemesis Yggralith Zero (x500000)"
-$mult = 500000
+name = "Respawn Global Nemesis Yggralith Zero (x-1)"
+$mult = -1
```
intra0, line 45: blade more tickets:
new x1 option, and changed the x500000 to x-1 (it was always a negative, 500000 in hex is 0x7a120, the preset can only be 16 bytes so it become 0xa120, which is a negative).
this name change will not affect anybody. when the game had online you *saw* when somebody was using this preset. it happened extriemly rarely and basically no one used it.
its main puprose of respawing the global nemesis server side has no use now with the online being dead.
you also wouldn't just leave this setting on casually.

so this name change affects no one, and I just want to mod to accurately describe what its doing instead of lying to the user about how it works, so changing it to a x-1 to reflect whats actually going on

src/XenobladeChroniclesX/Mods/CharacterLevel/patch_level.asm
```diff
@@ -31,24 +32,29 @@ _charLevelCheat:
 ; Character Level
 li r4, $level
 stb r4, $memoffset (r3)
+; Character exp
+exp = $exp
+lis r4, exp@ha
+addi r4, r4, exp@l
+stw r4, $memoffset+2 (r3)
```
intra0, line 39: new feature to the level change mod to change exp.
as well as being a new feature, also fixes some visual glitches with having above max EXP when becoming level 1

src/XenobladeChroniclesX/Mods/CharacterLevel/readme.txt
```diff
@@ -0,0 +1,247 @@
+this file is documenting all the data about a character that can be changed.
+fell free to use this document to help make your own character mods
+document made by googlebonker
```
intra0, line 3: new readme.txt document in the level change mod describing all the character data

intra0 (commented):
I made a review desribing all the changes and why I made them

intra0:
These are also the version's I've been using on my local machine for a few months at this point.
so all the changes are extremely stable.

src/XenobladeChroniclesX/Enhancements/FancyFX/rules.txt
```diff
@@ -143,36 +178,3 @@ width = 80
 height = 46
 formats = 0x816
 overwriteFormat = 0x820
-
-[Preset]
-name = 1080 -> 1440
-$width = 2560
-$height = 1440
```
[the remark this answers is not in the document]
getdls: FancyFX should probably be deprecated inst of incorporated. It's one of the earliest things I worked on, with a bunch of magic values that don't really scale."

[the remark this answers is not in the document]
intra0: yeah, idk anyone who actually uses it

intra0:
making new PR to solve the branch conflicts

PR #690, opened 2025-03-04
[XCX] [MK8] [DKC] Add credits comment | no features or code changes

intra0:
Really want the XCX credits comments in the repository by march 20th because of all the new eyes from the switch version
threw in mk8 and dkc just because.
will do this for every game at some point after the 20th


intra0:
would this commit be a good time to move the [DKC] mods into src or should I leave that out?
feel like I should leave it out

PR #693, opened 2025-04-01
[Splatoon] Superjump With The DPAD (Mod)

primaviera:
I ported the splatoon patches from SwapDRC to Cemu, instead of having to use your mouse to perform a superjump, you can just hold A and then press the D-Pad (it doesn't have to be in this order, but you might accidentally send a signal to your teammates if you do it the other way around).
I tested this several times and it works fine.


Crementif:
Yo, this looks great, thanks for your work!

PR #695, opened 2025-04-08
[XCX] clean up and add new features to mods - fixed

intra0:
Adds new features to the Enemy stats, max money, Global Nemesis offline, a x0 option for the exp mult (very useful in combination with the level change mod), a x0 & x1 for the other points mult,

implements suggested style changes from my last XCX PR

Removes unnecessary .int decorations because they bother me

Changes the name of "camera distance mod" to "default camera distance"
people will have to re-apply the mod and thats bad, but its old name is really missleading and it just bothered me a lot.

there may be some unnessisary changed to the .asm styling but I don't feel like reverting every file, and those style changes have at least 1 actual real change (not counting the pure deleting of unnecissary ints)

### DIFF
--- a/src/XenobladeChroniclesX/Enhancements/AntiAliasing/rules.txt
+++ b/src/XenobladeChroniclesX/Enhancements/AntiAliasing/rules.txt
@@ -9,9 +9,9 @@ version = 6
 [Default]
 $Preset:int = 0
 
-$subPix:float = 0.75
-$edgeThreshold:float = 0.166
-$edgeThresholdMin:float = 0.0833
+$subPix:double = 0.75
+$edgeThreshold:double = 0.166
+$edgeThresholdMin:double = 0.0833
 $FXAAQUALITYPRESET:int = 12
 $FXAAGREENASLUMA:int = 0
 $FXAADISCARD:int = 0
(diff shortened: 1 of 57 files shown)


intra0:
accidentially uploaded an outdated readme.md
gonna fix that real quick

intra0:
double checked everything with #689, and this PR is looking as good as that one :+1:

Crementif:
Hey, just cleaning up old PRs. Did we merge this with a later PR or did I forget about this... Hopefully not the latter, but it does happen. Let me know if it should still be merged?

intra0:
forgot about this one.
think everything here is good to be merged.
its just cleaning up extra modules that add an int that do nothing, and theres a few new features. (the enemy stats mod is the big one)
sorry for the late reply.

all of my PR's have the allow edit from maintainers turned on, so you can make any edits you want.
in the graphics packs 7 draft, the .asm patches have been ported to all versions, and the sonic forces quick boot mod has had major changes made to it (no longer need to manually edit a file)

I've made one more very simple XCX graphics pack I'll add to this PR and it should be good to merge

intra0:
image

PR #696, opened 2025-04-09
Imaginators Portal Fix

zephyrUkon:
This graphics pack mod will allow you to bypass the corrupt figure loop if you want to use Sanseis, Villains or Creation Crystals .sky files created with Cemu
I did not make the files to fix the Imaginators emulated portal.
I found it from here https://drive.google.com/file/d/1_Q8gcjfHACx9L9o640CxAka-9Hmccdew/view
I believe the person who made the files doesn't mind it being added to Cemus graphics pack collection because
he made a video on YT where he shared links showcasing the files. so all credit to https://www.youtube.com/@moltenlavacore

### DIFF
--- a/Imaginators Portal Fix
+++ b/Imaginators Portal Fix
@@ -1,0 +1,2 @@
+Added the file to fix the Skylanders Imaginators corrupt loop
+It is basicly the bypass NFC card but as a mod. wich can be anabled all the time


Crementif:
Your PR seems to have just added a file with a similar text, so it can't be merged as-is. But also, I don't really wanna add mods without the explicit permission from their creator. I'm closing the issue for that reason, but feel free to reopen a new PR that's fixed once you've asked for permission.

I would understand why the author of the mod would want to at least gain the views from the work he did, even if its probably not the biggest compensation, so I don't think your argument holds up about it being fine to upload because he also shared them on his personal platforms.

PR #702, opened 2025-04-23
[WWHD] Add 16:10 Steam Deck resolution with correct HUD and menus

depende3000:
Add support for 16:10 1280x800 Steam Deck's resolution. I changed the patches to the keep the 16:9 aspect ratio in the hud and menus to avoid cropping, that's why aspect ratio variables are splitted in two.

Here's a picture of it running on my deck:
IMG_8104

Edit 24th April, 2025:
3D World is corrected in 16:10 aspect ratio. HUD, and UI are a bit stretched because they are rendered in 16:9. If they were rendered in 16:10 they will be cropped due to engine limitations. Anyway, 16:9 HUD and UI distortion is minimal in 16:10 and is not annoying at all if tested on the Steam Deck's display :)

Also in the Steam Deck, remember to set using "Game mode" in Cemu "Stretch to window" aspect ratio and "bilinear" since the emulator believes it's being rendered in 16:9.

Happy to receive your feedback 👍

### DIFF
--- a/Resolutions/WindWakerHD_Resolution/rules.txt
+++ b/Resolutions/WindWakerHD_Resolution/rules.txt
@@ -2,7 +2,7 @@
 titleIds = 0005000010143400,0005000010143600,0005000010143500
 name = Resolution
 path = "The Legend of Zelda: The Wind Waker HD/Graphics/Resolution"
-description = Changes the resolution of the game. Made by getdls and Morph.
+description = Changes the resolution of the game. Made by getdls and Morph. 16:10 by Depende3k.
 version = 4
 
 [Preset]
@@ -13,7 +13,8 @@ $gameWidth = 1920
 $gameHeight = 1080
 $lightSource = 1.0
 $scaleShader = 1.0
-$aspectRatio = (16.0/9.0)
+$aspectRatioWorld = (16.0/9.0)
+$aspectRatioUI = (16.0/9.0)
 
 [Preset]
 name = 320x180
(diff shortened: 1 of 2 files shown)


Yoghurt4C:
lgtm

depende3000:
Thanks for the review! I'm not really sure if I need to move this into some kind of additional flow, or if this the right path and I just should wait for additional reviews or merge/approval. If you are familiar with this, any advice of help will be appreciated. Cheers!

Resolutions/WindWakerHD_Resolution/rules.txt
```diff
@@ -123,13 +134,36 @@ $gameWidth = 1920
...
+$lightSource  = 1.0
+$scaleShader = 1.0
+$aspectRatioWorld = (16.0/10.0)
+$aspectRatioUI = (16.0/9.0)
+
+[Preset]
+name = 1280x800 (16:10)
+$width      = 1280
+$height     = 800
+$gameWidth  = 1280
```
Crementif, line 161: So, you don't want to change the $gameWidth and $gameHeight. These are meant to be the game's original resolution aka 1920x1080. Then, the $width and $height are what the wanted resolution should be.

Below these presets, you can find the `[TextureRedefine]` group. This is where the resolution of the main screen texture (and many other texture resolutions) are changed to be rendering at the resolution you want. For this, it uses the formula: `overwriteWidth = ($width/$gameWidth) * 1920`. Since you've changed both of these, the division will be 1, and hence why its still rendering at 1920x1080. Its also why you needed to use the stretch option. If you change the main resolution (aka the 1920x1080 one) to 1280x800, it should automatically adapt to the entire gamepad.

Anyway, could you maybe try changing that and testing it? Ideally we don't need to have users change the scaling options, and the game should just work. I do like the separation of the UI and game world aspect ratio, and I think it could still help.

Crementif (commented):
Sorry for not being able to get earlier to your work. I hope my response helps and fix some of the issues you mentioned. It'd be a great game to run on a steam deck, so let me know if my suggestions help with that.

PR #712, opened 2025-06-11
[Lego City Undercover] Add shadow resolution

Nintendocustom:
grafik
grafik


Crementif:
Thanks! Sorry that it took a while to get merged!

PR #721, opened 2025-09-03
[TPHD] Add Cheats and more Screen Mirroring options

Ori-Jakob:
Changes:

- Add a bunch of various different cheats
  - Moon Jump
  - Always Great Spin Attack
  - Rupee Multiplier
  - Damage Multiplier
  - Never Decrease (Ammo)
  - Climb/Crawl Fast
  - Faster Rolling
  - Drop Duration Modifiers

- New Screen Mirroring GFX pack
This allows you to choose between mirror mode and normal mode, irrespective of the type of save you have (Normal or Hero mode). It even works on the fly while playing the game.


Crementif:
Seems like a great pack. The screen mirroring also mirrors the controls right? Like, it isn't literally mirroring the screen, and the input settings also require to be mirrored?

PR #734, opened 2026-01-09
[TTT2] Add "Reduce input lag" mod

WAZAAAAA0:
A mod that removes the forced "baked-in" input lag.
(same as its RPCS3 counterpart <https://wiki.rpcs3.net/index.php?title=Tekken_Tag_Tournament_2#Patches>)


Crementif:
Ah, cool! It is indeed questionable haha. I've also made your graphic pack be turned on by default, so anyone who updates it and plays it on Cemu will use your graphic pack by default. Thanks for porting it to Cemu!

PR #743, opened 2026-04-03
BotW: fix shadow cascade seam at 1280x800 (Steam Deck native resolution)

scndls:
## Summary

**Critical fix for Steam Deck users.** Eliminates the dark horizontal shadow band visible when running Breath of the Wild at **1280x800** — the Steam Deck's native resolution. This bug has forced Steam Deck users to run at 1440x900+ as a workaround, wasting GPU resources on handheld hardware where every frame counts.

Likely also fixes shadow artifacts at other non-16:9 resolutions (4:3, 21:9, 32:9) that share the same root cause.

### Before (1280x800, shadow seam visible)
Dark horizontal band across the terrain at shadow cascade boundaries, most visible when looking downhill.

### After (1280x800, same settings)
Clean shadows, no visible cascade seam.

## Root Cause

`TextureRedefine` height formulas for GPU-aligned textures scale the **alignment padding** along with the content height.

BotW's GPU creates render targets at aligned heights. For example, the half-res depth buffer is 640x368 (368 = 360 content + 8 bytes GPU alignment). The formula:

```
overwriteHeight = (($height/$gameHeight) * 368) + 0.49
```

At 1280x800 (`$height/$gameHeight = 1.111`), this produces **409**. But the non-aligned variant (640x360) correctly produces **400**. The 9-pixel mismatch (409 vs 400+8=408) between aligned and non-aligned versions of the same render target causes Cemu's texture operations to produce incorrect shadow cascade boundaries.

Additionally, **square textures** (64x64 depth stencil, 192x192 minimap, 512x512 heatwave) are scaled non-uniformly at non-16:9 aspect ratios (e.g., 64x64 becomes 64x71), distorting depth lookups used in shadow computation.

**Why 1280x800 specifically?** At this resolution, `$width/$gameWidth = 1.0` (width unchanged) while `$height/$gameHeight = 1.111` (height scaled). This creates a maximally non-uniform scale where width stays native but height stretches — the worst case for these formulas.

## Fix

**Aligned height formulas** — separate content scaling from alignment padding:
```
# Before (scales padding with content):
overwriteHeight = (($height/$gameHeight) * 368) + 0.49

# After (scales content, preserves padding):
overwriteHeight = ($height/$gameHeight) * 360 + 8
```

Same pattern applied at all mip levels (1/2, 1/3, 1/4, 1/6, 1/8, 1/16).

**Square textures** — use width ratio for both axes to stay square:
```
# Before (non-uniform stretch):
overwriteHeight = ($height/$gameHeight) * 64

# After (uniform):
overwriteHeight = ($width/$gameWidth) * 64
```

## 16:9 Compatibility

At native 720p (`$height/$gameHeight = 1.0`), all formulas produce **identical values** to the old formulas. Zero risk of regression at native resolution.

At upscaled 16:9 (e.g., 1920x1080), the half-res aligned height changes from 552 to 548 — this is actually more correct (proper aligned half of 1080 is 544+4=548).

## Test plan

- [x] 1280x800 (16:10, Steam Deck) — shadow seam eliminated, confirmed in gameplay
- [ ] 1280x720 (16:9, native) — verify no regression
- [ ] 1920x1080 (16:9, upscaled) — verify no regression
- [ ] 2560x1080 (21:9) — verify improvement

### DIFF
--- a/src/BreathOfTheWild/Graphics/rules.txt
+++ b/src/BreathOfTheWild/Graphics/rules.txt
@@ -666,7 +666,7 @@ height = 368
 formats = 0x001,0x005,0x007,0x019,0x01a,0x80e,0x806,0x816,0x820,0x41a
 # formatsExcluded = 0x431 # Exclude 0x431 which is used for adventure log images
 overwriteWidth = ($width/$gameWidth) * 640
-overwriteHeight = (($height/$gameHeight) * 368) + 0.49
+overwriteHeight = ($height/$gameHeight) * 360 + 8
 
 # Required 1/2 resolutions
 [TextureRedefine]
@@ -684,7 +684,7 @@ width = 384
 height = 192
 formats = 0x001 # World lighting - Red
 overwriteWidth = ($width/$gameWidth) * 384
-overwriteHeight = ($height/$gameHeight) * 192
+overwriteHeight = ($height/$gameHeight) * 180 + 12
 
 # Required 1/4 resolutions
 [TextureRedefine]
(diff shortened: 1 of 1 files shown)


scndls:
Pulling this back — the shadow seam reappears at higher resolutions (e.g. 1920x1200). The fix only holds at native 1280x800. Need to investigate further before resubmitting.

Issue #568, opened 2022-09-07
[BOTW] Aspect Ratio 16:10 Bug Steam Deck


mrtawsiti3:
Cemu Experimental 2.0.1
Steam Deck

Aspect Ratio 16:10: Shadow issue visible during the day, you can see it in front of Link
https://cdn.imageupload.workers.dev/j0bifsqR_P_20220907_113009.jpg

Aspect Ratio 16:9: How it should look
https://cdn.imageupload.workers.dev/JEDvLVWO_P_20220907_113330.jpg

ElieLiabeuf:
I have  the same issue. It seems that it doesn't affect all 16:10 res. If you use 1440x900 instead of 1280x800 there is no shadow lines.

ROGsamurai:
the deck should be able to handle 1920x1200. Twilight runs great with that

ASmyrnyi:
> I have the same issue. It seems that it doesn't affect all 16:10 res. If you use 1440x900 instead of 1280x800 there is no shadow lines.

thanks, works like a charm

scndls:
hey guys, I found the issue and provided the fix to the devs: https://github.com/cemu-project/cemu_graphic_packs/pull/744
now they only need to integrate it to their product. (or you are techy enough to fix it on your own by my description)

PR #745, opened 2026-04-07
[BotW] Add Graphics (16:10 Effect Fix) workaround for non-16:9 rain and shadow artifacts

scndls:
## What this fixes

At non-16:9 resolutions (16:10, 21:9, 4:3, 3:2, etc.), BotW exhibits two rendering artifacts that persist even with #744 applied:

1. **Rain bottom-strip** — when it's raining and the camera is pitched downward, a horizontal strip of broken rain particles appears at the bottom of the screen. Most visible on Steam Deck at native 1280×800 and on any 16:10 display at 1920×1200.
2. **Shadow cascade seam** — the pre-existing #568 bug, partially addressed by #744 but still visible at non-16:9 aspects.

Both artifacts disappear when the aspect ratio is switched to 16:9 in the Graphics pack, confirming the bugs are driven by how the depth-pyramid `TextureRedefine` entries scale at non-uniform aspect ratios.

## Root cause

The Graphics pack scales the entire depth pyramid proportionally with `$width/$gameWidth` and `$height/$gameHeight`. At 16:9 these ratios are equal (uniform scaling) so the depth pyramid stays proportionally consistent: main is N× the source, half-res is (N/2)×, quarter-res is (N/4)×, etc.

At non-16:9 the ratios diverge. For example at 1280×800:
- `$width/$gameWidth = 1.0`
- `$height/$gameHeight = 1.111`

This introduces a non-uniform Y stretch in every depth-pyramid downsample-chain `TextureRedefine`. BotW's rain pixel shader samples half-res depth using hardcoded native-720 assumptions, and when Cemu stretches the half-res depth buffer to match the upscaled main framebuffer, the rain shader's sampling near the bottom of the screen reads out-of-expected-range rows.

This was verified empirically by bisection of the Graphics pack's `TextureRedefine` entries on Steam Deck 1280×800:
- **0 depth-chain entries** (only main framebuffer upscaled): artifact-free
- **Half-res depth pair added** (`640×368` + `640×360`): rain strip appears
- Any further entries added: rain strip persists
- Modifying the half-res pair formulas to equalize the two buffers: introduces a shadow cascade seam (the 8-pixel padding gap is required for cascade alignment correctness)

We could not find a formula-level workaround. We also could not patch the rain pixel shader without identifying its hash (future work).

The only consistent, artifact-free configurations are:
1. All depth-pyramid entries scaled (current main pack — correct at 16:9 only)
2. Only the main framebuffer scaled, all downsample-chain entries left at native dimensions (this PR — correct at all aspects)

## The fix

Adds a new workaround pack at `src/BreathOfTheWild/Workarounds/GraphicsNon16By9/` that is a drop-in replacement for the main Graphics pack at non-16:9 aspect ratios:

- Identical `[Definition]` (modified name/path/description only), `[Default]` variables, all `[Preset]` blocks (aspect ratio, resolution, ultrawide HUD, AA, shadows, shadow draw distance)
- Identical `patch_AspectRatio.asm`, `patch_GUIAspectRatio.asm`, `patch_GUIScreens.asm`
- Identical shader patches (all 37 `*.vs.txt` / `*.ps.txt` files from `Graphics/`)
- **Only one rendering `TextureRedefine`**: the main framebuffer (1280×720 → user-selected resolution)
- Plus the five shadow cascade `TextureRedefine` entries (aspect-independent, driven by `$shadowRes`)
- All other depth/bloom/GUI/cubemap downsample `TextureRedefine` entries from `Graphics/rules.txt` are removed

Total: 41 files. The 40 non-`rules.txt` files are byte-identical copies of the corresponding files in `Graphics/`, so the pack is fully self-contained and tracks `Graphics/` for any future shader/asm updates that need to be re-mirrored.

Users at non-16:9 aspect ratios disable the main Graphics pack and enable this workaround pack instead. Users at 16:9 continue to use the main Graphics pack as before — **the main Graphics pack is unchanged by this PR**.

## Quality tradeoff

Non-16:9 users lose the depth-pyramid upscale chain for downsampled effects (SSAO, fog masks, bloom downsamples). At 1280×800 on Steam Deck (the most common use case for this fix), the difference is imperceptible — verified across multiple scenes and weather states.

## Testing

- **Steam Deck 1280×800 (native)**: rain strip gone, shadow cascade seam gone, general rendering flawless. Tested across multiple biomes (Hyrule Field, Akkala, Hateno), weather states (rain/clear), and camera angles (flat/steep downward).
- **1920×1200 16:10**: rain strip gone, shadow cascade seam gone.
- **16:9 users**: unaffected (main Graphics pack unchanged).

## Why not gate conditionally inside the existing Graphics pack

Cemu's `TextureRedefine` parser silently ignores `condition = ...` clauses inside `[TextureRedefine]` blocks (verified empirically with both an aspect-ratio condition and an always-false condition — neither prevented the entries from applying). It also does not evaluate complex formulas with user-defined integer variables in arithmetic expressions like `$useDepthPyramid * X + (1 - $useDepthPyramid) * Y` (verified with a `$useDepthPyramid:int = 0` toggle that had no effect on the rendered output). Without a working conditional mechanism inside `rules.txt`, a separate pack is the cleanest approach that preserves zero regression for 16:9 users.

## Related

- #568 — original BotW shadow cascade seam report
- #744 — open PR for the half-res rounding fix; this PR is **complementary**, not a replacement. #744 helps 16:9 users; this PR fixes the larger non-16:9 artifact class.

## File list

```
src/BreathOfTheWild/Workarounds/GraphicsNon16By9/
├── rules.txt                       (the actual fix — 669 lines, 6 TextureRedefines)
├── patch_AspectRatio.asm           (byte-identical copy from Graphics/)
├── patch_GUIAspectRatio.asm        (byte-identical copy)
├── patch_GUIScreens.asm            (byte-identical copy)
└── *.vs.txt / *.ps.txt             (37 shader patches, byte-identical copies)
```

### DIFF
--- a/src/BreathOfTheWild/Workarounds/GraphicsNon16By9/01ba1a725afa0b96_0000000000000000_vs.txt
+++ b/src/BreathOfTheWild/Workarounds/GraphicsNon16By9/01ba1a725afa0b96_0000000000000000_vs.txt
@@ -1,0 +1,138 @@
+#version 420
+#extension GL_ARB_texture_gather : enable
+#extension GL_ARB_separate_shader_objects : enable
+#ifdef VULKAN
+#define gl_VertexID gl_VertexIndex
+#define gl_InstanceID gl_InstanceIndex
+#define SET_POSITION(_v) gl_Position = _v; gl_Position.z = (gl_Position.z + gl_Position.w) / 2.0
+#else
+#define SET_POSITION(_v) gl_Position = _v
+#endif
+#ifdef VULKAN
+#define ATTR_LAYOUT(__vkSet, __location) layout(set = __vkSet, location = __location)
+#define UNIFORM_BUFFER_LAYOUT(__glLocation, __vkSet, __vkLocation) layout(set = __vkSet, binding = __vkLocation, std140)
+#define TEXTURE_LAYOUT(__glLocation, __vkSet, __vkLocation) layout(set = __vkSet, binding = __vkLocation)
+#define SET_POSITION(_v) gl_Position = _v; gl_Position.z = (gl_Position.z + gl_Position.w) / 2.0
+#define GET_FRAGCOORD() vec4(gl_FragCoord.xy*uf_fragCoordScale.xy,gl_FragCoord.zw)
+#define gl_VertexID gl_VertexIndex
+#define gl_InstanceID gl_InstanceIndex
+#else
+#define ATTR_LAYOUT(__vkSet, __location) layout(location = __location)
+#define UNIFORM_BUFFER_LAYOUT(__glLocation, __vkSet, __vkLocation) layout(binding = __glLocation, std140) 
+#define TEXTURE_LAYOUT(__glLocation, __vkSet, __vkLocation) layout(binding = __glLocation)
+#define SET_POSITION(_v) gl_Position = _v
+#define GET_FRAGCOORD() vec4(gl_FragCoord.xy*uf_fragCoordScale,gl_FragCoord.zw)
+#endif
+// This shader was automatically converted to be cross-compatible with Vulkan and OpenGL.
+
+
+const float resScale = $height/720.0;
+
+// Night star (size)
+
+// shader 01ba1a725afa0b96
+UNIFORM_BUFFER_LAYOUT(1, 0, 0) uniform uniformBlockVS1
+{
+vec4 uf_blockVS1[1024];
+};
+
+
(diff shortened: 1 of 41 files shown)


scndls:
Closing temporarily — pulling this back to verify the fix still holds at 1920x1200 after a setup conflict was discovered (both the original Graphics pack and this workaround were enabled simultaneously, which masked the test results). Will reopen once retested cleanly. Sorry for the noise.

PR #748, opened 2026-04-16
[MK8] Metal Shaders

emiyl:
Added are the shaders for the MK8 graphic packs, rewritten for Metal.

- [x] Graphics
- [x] Contrasty
- [ ] NoHud


Crementif:
Is this mergeable yet? Just wondering if progress was made. Would be a cool game for metal users, especially since it uses so many blurs.

PR #749, opened 2026-04-26
[WWHD] Add Wind Waker HD Cheats

Ori-Jakob:
List of available cheats:

- Moon Jump
- Sail Speed
- Swim Speed
- Save/Load Coordinates
- Teleport to Boat
- Automatic Wind Direction (always tailwind)
- Rupee Multiplier
- Rupees Never Decrease
- Remote Bombs
- Invincibility (No hurtbox)
- Infinite Cheats
    - Hearts
    - Magic
    - Swim Stamina
    - Arrows
    - Bombs
- Link Damage Taken Multiplier
- Link Damage Given Multiplier


src/WindWakerHD/Cheats/MegaCheats/patch_cheats.asm
```diff
@@ -0,0 +1,487 @@
+[MoonJump]
+moduleMatches = 0x475BD29F
+
+.origin = codecave
+
+_controlleraddy:
+    .int 0x4422e078
```
Exzap: Unfortunately hardcoding addresses to dynamically allocated memory is guaranteed to break with Cemu updates sooner or later. There is also no guarantee that the address will even be the same for different runs on the same Cemu version, since people can use stuff like [system RPLs](https://wiki.cemu.info/wiki/System_RPL_Files) which will shift all allocations.

You should see if you can find a pointer to the area in the data section of the RPX. Anything within .code and .text section has stable offsets

Ori-Jakob: Thank you for the heads up on this. I found a location in .data with a pointer to the controller.

Crementif:
The only issue I can really see is that this only works for US copies of the game. If I look at other patches of the game, I see that the addresses are quite different so I do think that we might want to add a disclaimer to the description or potentially port it.

But seems like you did some awesome work here, definitely appreciated!

Ori-Jakob:
> The only issue I can really see is that this only works for US copies of the game. If I look at other patches of the game, I see that the addresses are quite different so I do think that we might want to add a disclaimer to the description or potentially port it.
> 
> But seems like you did some awesome work here, definitely appreciated!

I should have tested the other versions instead of blindly adding all the Title IDs for the other versions.  I guess the EUR and JPN versions have some different code. For now, I can remove the other Title IDs from the rules.txt so they don't appear in the other versions graphic pack menus. TPHD didn't seem to have this problem though, so I wonder what makes WWHD so different?

Crementif:
The issue is that the moduleMatches = 0x475BD29F is equivalent to the WWHDUS version. You can see this in the https://github.com/cemu-project/cemu_graphic_packs/blob/master/Workarounds/WindWakerHD_IntelFixes/patches.txt for example. Nintendo compiled a different executable (aka .rpx file) for each region for this game, which usually means that the addresses are offset, usually. For some games this is a very small offset, but as you can see from the patches.txt file I linked it seems like there's quite a distance. And although you could blindly try that offset, I think its a bit too risky without testing.

As for the reason what is different inside the code... I think they probably just used the old gamecube code as-is to port the game. And back then region differences were probably more significant with PAL TV screens and such.

And yeah, getting rid of the title id is probably good enough. Then I'll merge it asap.

Ori-Jakob:
> The issue is that the moduleMatches = 0x475BD29F is equivalent to the WWHDUS version. You can see this in the https://github.com/cemu-project/cemu_graphic_packs/blob/master/Workarounds/WindWakerHD_IntelFixes/patches.txt for example. Nintendo compiled a different executable (aka .rpx file) for each region for this game, which usually means that the addresses are offset, usually. For some games this is a very small offset, but as you can see from the patches.txt file I linked it seems like there's quite a distance. And although you could blindly try that offset, I think its a bit too risky without testing.
> 
> As for the reason what is different inside the code... I think they probably just used the old gamecube code as-is to port the game. And back then region differences were probably more significant with PAL TV screens and such.
> 
> And yeah, getting rid of the title id is probably good enough. Then I'll merge it asap.

Ok, it has been restricted to US for now. What you said makes sense. I guess I will need to open both the EUR and JPN .rpx modules in Ghidra and try to find the correct memory offsets.

Crementif:
You want to work on that or would you rather merge it as-is?

Ori-Jakob:
~~Hmmm, I think merging as-is is fine for now. I am not sure how much time I will have to work on it, or how much time it will take to make the ports. However, I can always do that as a follow-up PR.~~

EDIT: Never mind, it was easier than I thought. I pretty much have the EU version done.

EDIT 2: Both EU and JP version now have support

Crementif:
Oh, that's awesome. Thanks!

Asia81:
Hello @Ori-Jakob , not sure this is the right place to report this issue, but in case :
Having "MegaCheats" activated, insta-kill Link when he take a single damage or dring any potion, despite having every cheats options on OFF

I recorded the issue : https://www.youtube.com/watch?v=ci_RooTNGo8

Thanks in advance if you can look at this.

Ori-Jakob:
> Hello @Ori-Jakob , not sure this is the right place to report this issue, but in case : Having "MegaCheats" activated, insta-kill Link when he take a single damage or dring any potion, despite having every cheats options on OFF
> 
> I recorded the issue : https://www.youtube.com/watch?v=ci_RooTNGo8
> 
> Thanks in advance if you can look at this.

I will take a look. I do notice that you are playing on the EU version of the game. I cannot replicated this on USA version which is what I use and it probably why this went by undetected. Since this PR is closed, it is better if you create an bug in the Issues tab and reference this PR number (`#` pound symbol followed by the PR number which is 749) in the bug issue itself

